PR #31, opened 2016-02-02
Frontpage prep and add columns classes

eepMoody:


source/index.rst
```diff
@@ -3,12 +3,47 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-Welcome to open5e
-==================================
+Open5e
+======
 
-*The best open source for the SRD, and the source of the best open content*
+Open5e aims to be the best open-source resource for fifth edition d20 content.
```
maxrothman: 5th Edition D&D content

source/index.rst
```diff
@@ -3,12 +3,47 @@
...
 
-Contents:
+.. rst-class:: hide-mobile
+
+What are we about?
+------------------
+
+.. rst-class:: hide-mobile
+
+In addition to an open, readable, and easily searchable resource for the 5e SRD, we also intend to provide:
```
maxrothman: This sounds like we already have it. Instead, maybe "We have lots of new things we'd like to add! At the top of our list are:"

source/index.rst
```diff
@@ -3,12 +3,47 @@
...
+
+.. rst-class:: hide-mobile
+
+* \• **Machine readable data** like monsters and spell lists, in formats like **csv, json, md, and xml**
```
maxrothman: Machine-readable

maxrothman:
After comments addressed, :+1: 

In the future it might be nice to try to only break between sections, but I think that is out-of-scope for this change.

source/index.rst
```diff
@@ -3,12 +3,47 @@
...
+* \• **Useful tools** such as **random tables, encounter calculators, treasure generators, And more...**
+* \• Plus the best **OGL community-created** content on the web!
```
maxrothman: This doesn't actually render as a list currently

eepMoody: This is how the theme is set up, that's why I added the bullets manually.

maxrothman: But it doesn't render as a list. Have you tried using an rst list?

source/index.rst
eepMoody: The term "D&D" is explicitly not part of the OGL per the included license. I'm using d20 because they don't provide an alternative.

maxrothman:
Does the SRD mention d20? I thought d20 referred specifically to 3.5

PR #85, opened 2016-02-07
Update header designs

eepMoody:
Updated the appearance and hierarchy of headers, and adjusted classes so they look correct.


maxrothman:
:+1:

PR #111, opened 2016-02-25
Add cache headers

maxrothman:
Relevant docs:
- https://docs.travis-ci.com/user/deployment/s3
- http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html#ExpirationDownloadDist


eepMoody:
I don't really understand this stuff beyond the basics, but assuming that this means "things will be cached for about an hour" then I'm good with that! 

:+1:

PR #114, opened 2016-03-09
added hotjar

jasonmcalpin:


jasonmcalpin:
That is HotJAR when spellcheck isn't messing with my spelling.

PR #126, opened 2016-03-21
Correct headings to match book & general cleanup

maxrothman:


maxrothman:
@eepMoody I think we might need some styling to make these changes look good. In particular:
* Table titles look unstyled and their anchor is overly-visible
* The first-letter highlight on smaller headers (h5 I think?) looks odd when there are multiple words in the title

eepMoody:
@maxrothman styling changes you mentioned have been made. Want to take a look and merge in if it's all good?

_themes/sass/_theme_rst.sass
```diff
@@ -174,19 +173,18 @@
 
 
   // This is the #href that shows up on hover. Sphinx's is terrible so I hack it away.
-  h1, h2, h3, h4, h5, h6, dl dt, p.caption
+  h1, h2, h3, h4, h5, h6, dl dt, p.caption, caption
```
maxrothman: Why both p.caption and caption? Don't you only need the latter?

maxrothman:
The h4 and anchors look good. I'm not crazy about the default styling on table titles though. Thoughts?

_themes/sass/_theme_rst.sass
eepMoody: I was avoiding changing the existing in case they style different types of captions differently. Would need to do a full review before I wanted to delete it everywhere.

eepMoody:
Yeah, I don't love it... let me do another more holistic styling pass as a separate issue, this one has lots of value in itself and I think it's worth going live as is?

maxrothman:
(Note for future me) Crap, there were conflicts in the css stuff, I have to rebuild the theme

maxrothman:
@eepMoody I think I fixed it?

eepMoody:
Let me just build that once I get to the office before you merge it in? There's a lot of font files that seen to have been changed or deleted, which doesn't seem like something the build should change...

maxrothman:
Take 2

eepMoody:
:+1:

PR #127, opened 2016-03-21
Max/crosslink combat

maxrothman:
NB: the link text will be Capitalized in all places where a condition is referenced. Is this ok?

Please give this a thorough review. There are some choices to be made about how/when/where we want to crosslink things. Especially keep an eye out for "Invisible". I *think* I caught all the places where it was linked erroneously...

### DIFF
--- a/source/Spellcasting/casting-a-spell.rst
+++ b/source/Spellcasting/casting-a-spell.rst
@@ -147,7 +147,7 @@ concentration. The following factors can break concentration:
    separate saving throw for each source of damage.
 
 -  **Being incapacitated or killed.** You lose concentration on a spell
-   if you are incapacitated or if you die.
+   if you are :ref:`srd:incapacitated` or if you die.
 
 The GM might also decide that certain environmental phenomena, such as a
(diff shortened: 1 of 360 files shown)


eepMoody:
Reviewed this for things like invisibility and others, haven't spotted any issues! I think the risk of some false-positive links is fine, given the huge benefit of linking these effects.

maxrothman:
Thoughts on having multiple links when a word is repeated? Following the "Wikipedia-style" would mean only linking the first one. Do we have a sense of when/how we want to repeat links? This came up in places where it would say something like this:

> This spell creates a 40-foot <sphere>. Nothing can pass through the sphere's wall.

(This is not a blocker for merging)

eepMoody:
Don't care about multiple links, as mentioned on #131. Is fine. Let's get this and that merged in before something causes more merge conflicts..

maxrothman:
For some reason, tests aren't showing up here after the rebase. [They passed](https://travis-ci.org/eepMoody/open5e/builds/120790571).

PR #131, opened 2016-03-21
Crosslink spell shapes

### LINKED ISSUE #91: Cross reference shapes (cone, cube, etc.) from spell descriptions

maxrothman:
Resolves #91


eepMoody:
SHIP IT

maxrothman:
Thoughts on having multiple links when a word is repeated? Following the "Wikipedia-style" would mean only linking the first one. Do we have a sense of when/how we want to repeat links? This came up in places where it would say something like this:

> This spell creates a 40-foot . Nothing can pass through the sphere's wall.

(This is not a blocker for merging)

eepMoody:
I'm fine with it showing up multiple times. Generally not going to cause a problem, and on very long pages it could be helpful.

PR #135, opened 2016-03-23
Fixes on the Magic Items.

Pvanduyse:
There was missing data, merged items, and incorrect formats that I
fixed. I also moved the Sentient Magic Items page from the Magic Items
folder to the Equipment folder, where It should be.


Pvanduyse:
Just fixed the error.

PR #140, opened 2016-04-02
Remove html suffix from all pages

maxrothman:
My thinking is that this makes it easier to get to pages directly by URL. Implications I could think of:
* cache-bust all users
* possibly negative SEO impact
* possibly break external links

Thoughts?


maxrothman:
Closed in favor of #141

PR #144, opened 2016-04-03
Fixes #106: Remove redundant alphabetical listing of monsters from toctree

### LINKED ISSUE #106: sidebar nav duplicates a-z list of monsters

eepMoody:


maxrothman:
Adding reference to issue from PR for future us: #106

source/monsters/index.rst
```diff
@@ -20,26 +20,16 @@ Monsters
...
-    :glob:
-    :maxdepth: 1
-
-    Monster Descriptions <monsters_a-z/index>
-
-------------------------------------------
-
-Monster Descriptions
-=====================
+.. container:: h1
```
maxrothman: Why this and not an actual header?

source/monsters/index.rst
```diff
@@ -20,26 +20,16 @@ Monsters
...
-
-      monsters_a-z/*/index
-      Miscellaneous Creatures <miscellaneous-creatures/index>
-      Nonplayer Characters <nonplayer-characters/index>
+        :name: monster-descs
+        :glob:
+        :titlesonly:
+        :maxdepth: 3
+
+        Monster Descriptions <monsters_a-z/index>
```
maxrothman: Why the explicit reference to the index rather than just a glob?

source/monsters/monsters_a-z/A/index.rst
```diff
@@ -5,5 +5,5 @@ A
    :titlesonly:
    :glob:
 
-   */index
+   **/index
```
maxrothman: Why this change?

Side note: we should really standardize how we do this. It's different in a bunch of places. We could probably just stick `**` in here and be good.

source/monsters/monsters_a-z/index.rst
```diff
@@ -1,12 +1,16 @@
...
+.. container:: columns
+
+    .. toctree::
+      :name: monsterlist
+      :titlesonly:
+      :glob:
+
+      */index
+      Miscellaneous Creatures <../miscellaneous-creatures/index>
+      Nonplayer Characters <../nonplayer-characters/index>
```
maxrothman: If we want these to appear here, do we want to just move them?

_themes/sass/_theme_type.sass
```diff
@@ -0,0 +1,236 @@
+// Typography
```
maxrothman: What does adding this file do?

eepMoody: Allows me to custom style the typography instead of having it be part of a bower dependency.

source/monsters/index.rst
eepMoody: If it was, it would appear in the TocTree and the bug wouldn't be fixed.

source/monsters/index.rst
eepMoody: If I did a glob of all indexes, they would appear in the toctree and the bug wouldn't be fixed.

source/monsters/monsters_a-z/A/index.rst
eepMoody: It's different depending on whether you want all subdirectories or just one, I believe? In this case I wanted all of them.

source/monsters/monsters_a-z/index.rst
eepMoody: We could, but it'd break outside links. I don't see any reason to just leave them as-is.

source/monsters/index.rst
eepMoody: To be more specific, I _think_ that this essentially references the structure included in that index file. If I globbed them, it would build a new instance of that structure, which would make it appear twice in the toctree.

source/monsters/monsters_a-z/A/index.rst
maxrothman: Right, but there's only 1 more level under A (it's Angels)

eepMoody: Ah, i see. It must have been from when i was trying to get it to render sub-indexes. Still never got it working.

source/monsters/monsters_a-z/index.rst
maxrothman: I agree that we shouldn't break outside links unnecessarily, but I also don't want that concern to become something that drags us down. If it's a strong concern we should figure out an automated way of addressing it, e.g. a better 404 page, auto-created redirects, etc. Part of the whole reason we chose Sphinx was that it allowed us to move things without breaking the site.

maxrothman:
Given the magnitude and visibility of this bug, :+1: regardless of the comments. We can address them later.

PR #160, opened 2016-06-13
Reorganize monster sections to flow better in ToC

maxrothman:
This change move Nonplayer Characters and Miscellaneous Monsters into
their own independed toctrees. This way, they'll appear as separate
sections in the main ToC and the Misc Monsters ToC will be able to flow
across columns between letter groups.

This change also introduces a weird sidebar nav bug: the moved sections
don't appear where they should when you're viewing their index page.
@eepMoody help?


eepMoody:
The bug about the sidebar is an expected behavior of the Sphinx ToC system. It's the reason they were grouped the way they are, so unless we come up with some way to get around the ToC system (eg. crawl the contents without actually creating a ToC entry) we're either going to have a broken nav or a messy page. I feel like the latter is something I'm more comfortable with...

PR #161, opened 2016-06-20
Consistency between "*Hit:*" and "*Hit*:"

snan:
Some monsters said "*Hit:*" and some said "*Hit*:" and I kinda prefer the former but there were way more of the latter so I changed the outliers to match the majority.
TL;DR:
It now always says "*Hit*:" and never "*Hit:* and this goes for Miss too.


maxrothman:
Thanks for the pull request! I'm 👍, and if nobody objects, I'll merge this shortly.

PR #162, opened 2016-06-20
Update to SRD 5.1

maxrothman:
I've worked through the complete diff, so this is ready for review!


thomax:
Looks very good to me! Tested make, and pages are built correctly, as far as I can tell. Good work :)

eepMoody:
👍

PR #163, opened 2016-06-21
Flesh Golem code block formatting

snan:
I'm iffy that this patch is a lone paragraph instead of a continuation of **Berserk** but it's better than how it is now, where the golem is a code block. Look how weird it looks on the web page: 
https://open5e.com/monsters/monsters_a-z/G/golems/flesh-golem.html


snan:
Uh, I just started using github (not new to git though). Turns out this was both of my patches in one? (The one that was already accepted and then this) How do I send just one?

PR #164, opened 2016-06-21
Missing newlines

snan:


maxrothman:
Thanks again!

By the way, if you're going to be doing a bunch of cleanup, feel free to lump commits together in the same PR. We care more about the atomicity of commits than of PRs.

snan:
Oh, can you apply a PR on commit-by-commit basis?
Still new to Github even though I've used Git before.

BTW I was working on machine-readable monsters (no promise that I'll ever see that project through), that's why I stumbled upon these weird spots, my RST reader gave me weird output there. (I'm not going through the RST files by hand except to double check when there's weirdness.)

I'm using a very crude custom pandoc plugin to output s-expressions (pandoc can read RST), then I'm doing a lot of cleanup and parsing of those s-expressions in Scheme. When I'm done I can make JSON, XML etc.

Here's my progress so far, obviously this isn't anywhere near usable yet, but just as a sneak peek. I haven't even started on the special abilities, the actions or the legendary actions yet. I'm doing the low hanging fruit such as ability scores, xp value etc first.
http://idiomdrottning.org/monsta.scm.txt <- I'm not creating this by hand, I'm trying to keep the entire chain automated from the RST files.

maxrothman:
A PR is on a branch to a branch, so if you add more commits to the branch more commits are added to the PR.

Hey, I'm working on machine-readable everything! You should send me an email (on my profile) so we can coordinate!

snan:
Right, but what if there are some commits in the branch that you do like and some that you don't like?

maxrothman:
Ah, yeah, you can't PR individual commits. In that case I usually make a new branch with just the commits I want on it.

PR #165, opened 2016-06-21
Why is this a code block?

snan:
This is a new version of the pull request that only has the one commit. I'll try to delete/cancel the other pull request. I got some help with github


maxrothman:
That looks like an artifact of the original PDF -> RST process. Good catch!

Issue #171, opened 2016-10-02
SRD Data


bgotink:
I've been looking around on the web, and I couldn't find any good sources for the 5e SRD data. I'm not interested in a PDF or HTML, I need data files (e.g. JSON or YAML) containing the spells, feats etc. in a structured way.

I've tried the `make json`, but outputs the files in a way that can be used in a website. I would like something similar to (example in YAML because it's easier to write than JSON)

``` yaml
name: Acid Arrow
level: 2
school: evocation
ritual: false
casting time: 1 action
range: 90 ft
components:
  verbal: true
  somatic: true
  material: powdered rhubarb leaf and an adder's stomach
duration: instantaneous
description: '''
A shimmering green arrow streaks toward a target within range and bursts
in a spray of acid. Make a ranged spell attack against the target. On a
hit, the target takes 4d4 acid damage immediately and 2d4 acid damage at
the end of its next turn. On a miss, the arrow splashes the target with
acid for half as much of the initial damage and no damage at the end of
its next turn.
'''
at higher levels: '''
When you cast this spell using a spell slot of 3rd
level or higher, the damage (both initial and later) increases by 1d4
for each slot level above 2nd.
'''
```

This would allow for easy filtering on school, level, components, duration etc.

Is there any interest to implement something like this? In my opinion it would not only be a great addition to the open5e website if we could filter dynamically instead of only on first letter and on class, but it would allow for other projects to use this project as datasource (by publishing as npm module?).

eepMoody:
Yeah, this is absolutely something we were hoping to do. Issue we ran into was figuring out how to parse the data we have into JSON or similar. (Once we get it into some sort of structured data, we figure it'll be easy to move into another type)

So far we were running up against issues with the Sphinx build process, but I've been debating moving our core build process out of Sphinx and basing our data elsewhere. (eg. constructing our Sphinx files from JSON snippets or similar)

Dream would be to have one source of data truth and build everything else from it, either from .rst > JSON or vice-versa.

If you have any thoughts, I'd be super interested to hear them!

maxrothman:
[...]

PR #172, opened 2016-10-20
Add utils/extract-spell-list.py

bgotink:
This python script parses spell data from the doctree elements according to what was discussed in #171. Note that the meta and link keys are not yet supported.

This is my first try at Python, so don't shoot me if what I wrote is not idiomatic 😉 

Note 1: I had to install sphinx and add the installation to my `PYTHONPATH` because the pickled documents contain sphinx-specific objects. (`export PYTHONPATH=/usr/local/opt/sphinx-doc/libexec/lib/python2.7/site-packages`)

Note 2: The script currently requires that it is run from the repository root, e.g. `utils/extract-spell-list.py | pbcopy`

The output can be found [here (gist)](https://gist.github.com/bgotink/429533f2b97c3fa97483b4fbff2bffd8)


eepMoody:
Hey @bgotink I'm trying to get this moving again and resolve all the pull requests I let sit waaay too long. I know there were some side conversations about this, do you remember if this left off in a good place? It looks like it isn't causing any issues merge-wise.

cameronblandford:
Since the code-base is heading in a different direction now, would everyone feel comfortable closing this PR? @bgotink @eepMoody

eepMoody:
I would be cool closing it out, even though it's a shame we can't use it... .rst is just so hard to use elsewhere.

PR #176, opened 2016-11-18
Add spells JSON file

tjcarroll11:
Added spell data as in JSON format


eepMoody:
Hi tjcarroll. Just wondering where you got this json file? We want to make sure any data we provide is matched with and built from the data we provide in text form (which is why we've not released something yet). If it's built from our source and we can repeatably run it, I think we can use this!

tjcarroll11:
I've added to my PR the cleanup of your MD file along with the python script that generated it.

eepMoody:
@tjcarroll11  Sorry about the slow on this. I had to do some travel and haven't had a chance to actually run the build and check out how it looks. I'm gonna get to it this week!

tjcarroll11:
Ha so I actually hadn't really looked at your repo (whoops). I just happened to have written the script in python! I'll check out your build this week and actually integrate my spell JSON generator script.

cameronblandford:
Not that it actually concerns me but could this branch get closed or re-examined?

PR #166, opened 2016-06-21
Monster's spell lists as lists + small bulette fix

snan:
So two separate things:
Monster's/NPC's spell lists are formatted in many ways across the various RST files. I want them to be all the same way. Blockquote (as some, not all, of the NPCs were), that's fine. Just as if they were any other special trait (as the monsters and a few of the NPCs were), that's a bit harder to work with but not impossible.
But I put them all in lists: none of the other monsters use lists and spell lists are just that, lists.
CSS will need to be tweaked perhaps.
Githubs RST preview doesn't show these nested lists so well, but they're there.

And the other thing was super minor, the Bulette had accidentally become line-quoted.

PR #192, opened 2017-01-31
Spaces added between fields to bulette

oganm:
Should have included this into #191. This also deals with the bulette fix from #166


oganm:
This is what it currently looks like. Bigly sad!
selection_054

PR #201, opened 2017-03-20
Update monk.rst

mstashev:
Fixing for pull requests 197, 199-200


eepMoody:
So this would be great as we'd love to include more, but unfortunately this content isn't in the wizards srd, which means we can't include it here without violating their copyright and leaving ourselves open to a DMCA takedown. :(

PR #216, opened 2018-01-15
Add Open Game Content from Kobold Press Midgard Heroes book

RealDeuce:
There's some issue with bullet lists, and I had to do terrible things to the template to support a different license.


RealDeuce:
Incorporating all previous requests into one big one.

PR #217, opened 2018-01-15
Split races into separate pages

RealDeuce:
This splits races into a one per page like monsters and spells.  This is likely the better format with the addition of other OGL races.


RealDeuce:
So I actually have the Midgard and Southlands heroes races done this way, so if you like this style, I could abandon #216.

RealDeuce:
Using this method for spells and monsters however doesn't seem to be a good fit... maybe a separate per-source list as well as a massive combined list for those?  I'm not really sure, but since I haven't finished the Tome of Beasts stuff yet, and haven't started the Deep Magic things, it's not vital yet.

RealDeuce:
Incorporating all previous requests into one big one.

PR #219, opened 2018-01-17
Kobold Press 5e races and monsters

RealDeuce:
Add races from Midgard Heroes and Southlands Heroes, and monsters from Tome of Beasts.

Split races into separate files to help this out.

Apparently Travis needs to run on *every* commit, so just squash it all.


eepMoody:
Taking a look now, this is a big set of changes! Thanks so much for taking a crack at this. :)

RealDeuce:
Heh, I have pretty much all the Kobold Press stuff, so I may peck away at various other player bits as they get used in my campaign.  The Deep Magic series in particular.

I wasn't sure if I should keep the non-SRD stuff separate (as I did) or mix it in with the SRD stuff and label it somehow, so I just went with what was easiest.  It likely won't work well for smaller sources like Deep Magic, or the even smaller sources like their Warlock series or the OGL stuff from their modules and blog, so before I start digging into them, some "right way" needs to be figured out.

I'm also not sure how/if to call out where Product Identity has been removed.  All the Arch-Devils, Demon Lords, and the Fey Lords and Ladies for example aren't included since they have proper names, and are characters... so are Product Identity.  Some of the monster descriptions had references to locations, plots, and story elements from Midgard so had those bits removed or rewritten... should those types of changes have a sidebar?

The only sidebars I included in this pull request are ones that are a mechanic that's needed to use the creature, but are Product Identity and couldn't be included... vanisher hats for the Scheznyki and the Arcane Wasting disease of the White Apes.

As far as linking conditions and spells, I only linked them if they were imposed or cast (or were used to describe some effect that was triggered).  I didn't link them from immunities and such.  I like it better that way, but it seems that for the SRD stuff, they're linked everywhere.  I change fix that if you want, I just like my way better.  :-)

eepMoody:
This is awesome. I don't think we need to worry about the Product Identity stuff, I included that general "Wandering Monsters?" on the index page for exactly this reason. ;)

It's actually a huge pain, since we can't even refer to the IP we're excluding without infringing on it. Yay copyright.

I think that down the road I would like to set up a tagging system so we can mix everything in, but this is an awesome way to get it started. I adjusted the titles slightly so that they fit a bit better with future plans I've got, but otherwise we're good! I'm going to merge this in as soon as everything passes and I get a chance to preview it.

PR #220, opened 2018-01-20
Add Deep Magic: Dragon Magic stuff.

RealDeuce:
This is sort of a trial of how to include smaller 3rd party stuff...

I've added a Third Party Content top-level item.

Spells are included in the A-Z indexes after the SRD ones.  In the letter-specific index, they have a caption, but in the "Spells A-Z" index, they are not separated from the SRD spells except the are at the end.

Feats are weird.  I'm not happy with the TOC generated in the characters section, but I'm not sure what to do about it.


eepMoody:
You've made an excellent point here. I think this might be a good opportunity to look into what sort of tagging and filtering systems I can get out of Sphinx. The "Third Party" section will get unwieldy over time, and I would prefer to keep the number of third-party sources limited (to stuff that's widely-used and accepted as balanced) but that'll still grow over time.

I'm going to hold off on this while I look into some options. If you have any ideas, feel free to email me and we can talk it over in more detail.

PR #221, opened 2018-02-07
Make "Interacting with Objects Around You" it's own section

WillMorrison:
Doesn't look like it belongs to the "Being Prone" section.


eepMoody:
Thanks for the catch!

PR #225, opened 2018-02-26
Issue 212

Athaman:
I've gone through and fixed a lot of the requested updates in this issue. The ones towards the end the mention needing to be updated in all files I haven't done yet but could probably manage through some basic grepping or seding. Let me know if you'd like them changed across the board.


eepMoody:
Thanks @Athaman. 

If you want to take a shot at grepping those general changes for italicizing the monster names, go for it, but this was super helpful already!

PR #244, opened 2018-08-12
local environment variables

Taormina:
Hoping to address https://github.com/eepMoody/open5e/issues/240. I was wondering about the `sys.path` setting and whether that was being used for anything else apart from the relative file location (because I'm doing it with an absolute location now, so if that parameter doesn't matter, we can just delete all of that). If its being used for something else, then we'll need to test/edit this more.

### DIFF
--- a/server/scripts/add_srd_monsters.py
+++ b/server/scripts/add_srd_monsters.py
@@ -1,5 +1,6 @@
 import os, sys
-sys.path.append('/Users/ean/dev/open5e/server')
+root_path = os.environ['OPEN_5E_ROOT']
+sys.path.append(root_path)
 os.environ['DJANGO_SETTINGS_MODULE'] = 'server.settings'
 import django
 django.setup()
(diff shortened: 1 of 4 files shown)


Taormina:
Also, I can see *that* I am failing the build, but not *how*. Am I missing something or is the build meaningless with the refactor in progress?

eepMoody:
@Taormina Don't worry about the failing build, it's still trying to run the old static stuff. We haven't actually set up the new build script, and it's going to be a bit of a rough transition. :)

I'll take a look through this tonight.

server/scripts/add_srd_monsters.py
```diff
@@ -1,13 +1,16 @@
 import os, sys
-sys.path.append('/Users/ean/dev/open5e/server')
```
eepMoody: It is great to see my local file system vanish from this repo! 😁

eepMoody:
Nuked my repo and did a fresh setup from this, worked perfectly. Thanks a bunch for the help!

PR #249, opened 2018-08-22
WIP:Json model import

eepMoody:
Taking a stab at building out models for this. If you guys think this looks good I'll start on some importer scripts next!

In particular, not sure that a ForeignKey makes sense for the many-to-one relationship going on with races/classes and their children?  Did I put them in the ideal location?


augustjohnson:
In my mind the reference between a CharClass and a Subtype was at the Subtype level.  So the foreignkey was in THAT table, rather than the class table, since each Subtype can only be tied to a single CharClass.  Then it's a One to One, making things a bit easier than a One to Many.

Other than that, this looks GREAT, and you may have had a good reason to relate them downwards.  I can build out the serializers if this seems to be working.

eepMoody:
@augustjohnson I really didn't, I was doing some research and people seemed split on it. Your logic makes sense, and it would make the relationship simpler!

One thing I could use an eye on: The import script I wrote is having a strange issue and I can't for the life of me parse out why. It's adding an extra blank row to the database before each actual row.

Nothing extra in the JSON, the script doesn't seem to have an extra loop or object creation in it. I'm totally stumped.

server/api/models.py
```diff
@@ -51,7 +55,78 @@ class Spell(models.Model):
...
+    age = models.TextField()
+    alignment = models.TextField()
+    size = models.TextField()
+    speed = JSONField()
+    speed_desc = models.TextField()
+    languages = models.TextField()
+    vision = models.TextField()
+    traits = models.TextField()
+    subtypes  = models.TextField()
+    sub_race = models.ForeignKey(SubRace, on_delete=models.CASCADE, null=True)
```
cameronblandford: Similar to the Archetype vs Character Class discussion, this foreign key should be on the sub-race. i.e. when you have a tree and you can only store one pointer on each node, you wanna use that pointer to define the parent :)

server/api/models.py
```diff
@@ -1,8 +1,11 @@
+import uuid
 from django.db import models
+from django.contrib.postgres.fields import JSONField
```
cameronblandford: are we deciding on postgres here?

server/api/models.py
```diff
@@ -51,7 +55,78 @@ class Spell(models.Model):
...
+    prof_weapons = models.TextField()
+    prof_tools = models.TextField()
+    prof_saving_throws = models.TextField()
+    prof_skills = models.TextField()
+    equipment = models.TextField()
+    table = models.TextField()
+    desc = models.TextField()
+    spellcasting_ability = models.TextField()
+    subtypes_name = models.TextField()
+    archetype = models.ForeignKey(Archetype, on_delete=models.CASCADE, null=True)
```
cameronblandford: If there are going to be situations where a charclass has more than one archetype, then the relationship should be stored on the archetype - doing it this way, there can only be one archetype per charclass (since the foreign key is a single int).

server/api/models.py
```diff
@@ -51,7 +55,78 @@ class Spell(models.Model):
...
+    speed = JSONField()
+    speed_desc = models.TextField()
+    languages = models.TextField()
+    vision = models.TextField()
+    traits = models.TextField()
+    subtypes  = models.TextField()
+    sub_race = models.ForeignKey(SubRace, on_delete=models.CASCADE, null=True)
+
+class Plane(models.Model):
+    slug = models.SlugField(unique=True, default=uuid.uuid1)
```
cameronblandford: It might make sense to do some refactoring in the models, since basically everything is using a slug, a name, and a desc. you can move that all into a parent class that these classes inherit from and save a few dozen repeated lines :)

cameronblandford:
@eepMoody re: being stumped, first fix the issue with the archetype stuff and see if that helps - there might be some undefined behavior when you put a non-PK into a field that only accepts a PK, and then save the model like what's currently happening.

If that doesn't help, try pip installing `ipdb`, then adding `import ipdb; ipdb.set_trace()` wherever you wanna throw a debugger and it'll let you step through the code! [Here's](http://www.christianlong.com/blog/ipdb-the-ipython-debugger.html) a good tutorial for it.

server/scripts/add_srd_class.py
```diff
@@ -0,0 +1,51 @@
...
+            c.table = f['table']
+        if 'desc' in f:
+            c.desc = f['desc']
+        if 'spellcasting-ability' in f:
+            c.spellcasting_ability = f['spellcasting-ability']
+        if 'subtypes-name' in cClass:
+            c.subtypes_name = cClass['subtypes-name']
+        if 'archetype' in cClass:
+            c.archetype = cClass['archetype']
+        c.save()
```
cameronblandford: After you save/insert the CharClass instance, you're going to want to loop through the `c.subtypes` in each entry, instantiating, populating, and saving those (with `c.slug` as their `parent_class`/`char_class`)

server/scripts/add_srd_class.py
```diff
@@ -0,0 +1,51 @@
...
+            c.equipment = f['equipment']
+        if 'table' in f:
+            c.table = f['table']
+        if 'desc' in f:
+            c.desc = f['desc']
+        if 'spellcasting-ability' in f:
+            c.spellcasting_ability = f['spellcasting-ability']
+        if 'subtypes-name' in cClass:
+            c.subtypes_name = cClass['subtypes-name']
+        if 'archetype' in cClass:
```
cameronblandford: Again, this looks a lil fishy to me - if I'm understanding correctly what's intended here (and I may not be), these two lines should be after the .save() call and should look something like:
```
if 'subtypes' in cClass:
    for subtype in cClass['subtypes']:
        # create and save new subtype record with relevant data
```

server/api/models.py
augustjohnson: As a note to this, perhaps we should agree on a few universal fields to add to the parent class?  Some metadata as well (like my document source)?
1. Name - (E.G. "Barbarian" or "Blinded")
2. Slug - ("magic-missile")  Consistent, for the URL.
3. Description - The Body of the document.  Lots of differences here, but here's a question:
Should the Description contain the Name at the top?  I was inconsistent here,  For things like Subtype I did not, for things like Conditions I did.
4. Meta_Source Work (this is a reference back to some SRD object sitting in a table somewhere.
Here's some questionable shared fields:
5. Meta_Page  (This would be the PHB, DMG, or MM page.)
6. Meta_Section or Meta_Parent or something... this is a way to organize the pages into navigation, although some of the sections kind of defy organization.

server/api/models.py
augustjohnson: Honestly, I kind of like the idea of being able to swap it out a few different ways.  Is it possible to keep it database platform agnostic?
https://docs.djangoproject.com/en/2.1/ref/contrib/postgres/fields/#django.contrib.postgres.fields.JSONField
According to this, the JSONField used here allows for indexing... but I'm not sure that's a big deal on something as small as speed.

cameronblandford: @augustjohnson the way to keep it platform agnostic would be to not use any database-specific fields. Anything usable on sqlite *should* also be platform agnostic because it has such a minimal feature-set. Postgres just gives you a bunch of extra tooling. Storing the JSON as a string would be a universally available option

server/api/models.py
cameronblandford: I like these! Along with the obligatory `last_edited` and `date_created` fields of course. And as for the description - I'd recommend not including the name - that way we can compute a description block that includes the name if we want to, but either way we won't have it listed twice.

eepMoody: Yeah, agree on leaving the name out of the description. Generally speaking you'd want to be able to populate that in a tooltip or other display, probably referenced by the name.

Agree on all those universal fields

PR #263, opened 2018-09-12
"Font of" -> "Font of Magic"

enkryptor:


eepMoody:
It appears the newest version of sphinx is broken. I'm trying to figure out what to do about it.

eepMoody:
Yeah, I've been tinkering away at this and been unable to find a fix. 

It looks like Sphinx has updated in such a way that one of its processes cannot be run on our host. Given that we're like 90% of the way through launching a new version (and how annoying Sphinx has been thus far) I think it might be better to put the time into that instead of trying to find a workaround or new host.

PR #272, opened 2018-10-12
Kubernetes and Docker Helper Script

acidDrain:
## Kubernetes

We containerized the Django API last week, and this week have been working on deploying to a cloud platform.

Last night, we built the manifests and deployed to a Google Cloud Kubernetes instance (my personal instance). This was really done out of curiosity and having an excuse to tinker in GCP.

I had to make tweaks to some of the Django and Docker config while I was building and tesitng the Kubernetes manifests/config. I ran into a few issues around environment variables and Django / Docker config that I started thinking about how to solve for others in the future.

Here's the deal: Django expects these configuration items (passed as environment variables):

* **[SECRET_KEY](https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-SECRET_KEY)** - the `SECRET_KEY` is used for session signing. If this is improperly configured or missing, the API would be exposed to potential privilege escalation and remote code execution vulnerabilities.
* **[ALLOWED_HOSTS](https://docs.djangoproject.com/en/2.1/ref/settings/#allowed-hosts)** - Django checks the `HOST` header of incoming HTTP requests against the `ALLOWED_HOSTS` configuration setting to prevent [HTTP Host Header Attacks](https://docs.djangoproject.com/en/2.1/topics/security/#host-headers-virtual-hosting). The `HOST` variable would normally be configured as the DNS/hostname of the system running Django.
* **[DEBUG](https://docs.djangoproject.com/en/2.1/ref/settings/#debug)** - Django shouldn't be run in production/publicly accessible with `DEBUG` = `TRUE`, as it could expose sensitive information, configuration flaws that could be exploited, or cause performance issues.

I thought it would be useful to parameterize the Docker build process (and execution) so that these values could be passed in based on the environment and need (e.g. troubleshooting).

To abstract some of the [technical details that Docker needs to be configured and run with](https://docs.docker.com/engine/reference/commandline/build/), I wrote a build script in bash. This led me to three questions that I wanted to discuss and get opinions on:

1. Is a helper shell script for Docker out of place in this repo, instead configure/store on the build system? If a script like this is OK, is a **`bash`** shell script acceptable, or should it be in Python to be more platform agnostic in case there are people contributing from Windows machines?
2. Should we consider containerizing other components of this repo/app? If so, I can work on coming up with a directory / file structure to support that (from the Dockerfile/build perspective).
3. Should we discuss secrets management process or standard so that a user can configure the variables mentioned above and at the same time avoid exposure? The pattern/process I'm used to, and started setting up, was to add a gitignored file in the root of the project called `.env` that contains a `KEY` = `VALUE` pair corresponding to the variables you want to configure. Those values would get injected into the container at build argument, and can also configure them as environment variables in the container image so they could be changed at run time if needed.

I can make these decisions on my own if everyone's OK with that, but I wanted to make sure there was at least opportunity for discussions and opinions.

Also, you can check it out [running on Kubernetes in GCP here](http://open5e-api.elasticplayground.com/) if you're curious.

@eepMoody @augustjohnson


augustjohnson:
Dude, huge changes, thanks a ton on this.
Regarding your questions:
1. I have no problems with bash.  Not a big deal either way.  It looks simple enough that if someone one windows wanted to recreate they probably could.

2. I think the way things are shaking out how are this:
Containerized deployment for the API and all supporting infrastructure.  Maybe on google?  Whatever.
SSL and caching offloaded to cloudflare free edition.
Vue static frontend hosted on something like netlifly.

3. I like the .env approach on secrets.

eepMoody:
I'll be honest that I'm only following about 75% of this, but that looks incredible! So my understanding from working with similar build setups is that I'd set up those environment variables (including the secret key) in DO or AWS, whatever my CI/CD system is?

To your points:
1. +1 to bash being fine. You can run bash on windows, and I feel like anyone doing Python/JS development on windows will know how at this point.
2. I'm not sure that's necessary at this stage, but I'm fine with it if someone wants to tackle it! Getting the UI layer and API set up within a single virtual environment would probably make local development simpler.
3. That sounds familiar to me. For convenient local development, I've seen a pattern of creating a .env.example file for them to copy that has all the appropriate settings for local development but no key. Would that work well here?

To points 2 and 3: is there anything I or someone else needs to do to let the UI layer access the API in a predictable way for local development?

PR #279, opened 2018-10-27
Navbar Collapse

christopherfowers:
This is just some initial work. There's not much to this at all. Currently just determines what to hide in the navigation menu based on current route.


eepMoody:
This is super slick man. I know you said you were going to be working on making it work before one has navigated, but I'm gonna merge this in tonight just so I can get it in the (increasingly upcoming) beta launch!

PR #280, opened 2018-10-30
Navbar Collapse and Docker-Compose mods

christopherfowers:
Hopefully this will work for the service and nginx deployment. 

Also replacing PR #279 with this.


christopherfowers:
Sorry about using multiple accounts here. Can make this confusing. Working to remedy that issue.

PR #286, opened 2019-01-31
WIP:Tome of beasts

eepMoody:
Add tome of beasts JSON & update importer scripts to parse it into the DB.


augustjohnson:
Your rewrite of my importer is highly encouraged :).

eepMoody:
@augustjohnson The changes are actually fairly minimal! It just looks like a lot because of whitespace and variable names.

I switched to a copy while doing the loop through abilities since some of the items need to be altered or deleted which was throwing errors for me.

Then in the main importer I just gated everything behind ifs so we can import documents that contain any subset of content that was in the original.

eepMoody:
Did this on the API side instead, which is already merged in!

Issue #264, opened 2018-09-21
Put spells and monsters into filterable and sortable tables


eepMoody:
Should be filterable and sortable by common attributes like casting time, school, type, duration, etc.

Bonus points: filter and sort by casting class so we don't need to make separate pages for those.

Bonus Bonus: Find a way to get some short summaries in there for the spells.

ScottBWar:
Opened a PR for the monsters table. Same way as the spells.

https://github.com/eepMoody/open5e/pull/293

eepMoody:
This was done by ScottBWar but I forgot to close the issue!

Issue #235, opened 2018-08-11
Adjust/regenerate Spells to include Classes as an array


eepMoody:
Currently this is only available as a string `"wizard, druid, cleric"`. It'd be great if it was an array of strings instead. `["wizard", "druid", "cleric"]` so that it can be better used for filtering, and parsed in cleverer ways on the frontend.

eepMoody:
Same goes for circles, archetypes, etc.

McCrossin:
Hey, I am looking for an issue to start off with.
Mind guiding me on how I can start out on this one?
Many thanks

augustjohnson:
@eepMoody is this one you're looking to solve in the API layer?  If so, we should transfer this issue to that project.

eepMoody:
Good point! I will go through and do so for all the API-related issues. They were made mostly when everything was one repo.

Tasups:
Wondering if I could take it on, if it's still open, and which file is it located in.

augustjohnson:
@Tasups This is still an issue, but I'm pretty close to solving it over in the API.  If you're interested in contributing some python, I can definitely find something.

augustjohnson:
See a big model upgrade here: https://github.com/open5e/open5e-api/tree/Spell-Model-Upgrade

The file you'd likely be adjusting if you were to take this on, would be https://github.com/open5e/open5e-api/blob/Spell-Model-Upgrade/api/models/spell.py

PR #289, opened 2019-02-04
Table View for Spell List Index

ScottBWar:
As suggested by your issue here: 
 #264 

A table view for spells, that can be sorted/filtered/viewed by other properties than name alphabetically.  It will make sense to add sorting by class after #235 is changed to give spell classes as arrays from the API


ScottBWar:
I guess I could change it to scrap the original list and just have this be the base spell list. Planning on adding pagination and some way to organize by class when I have time as well

eepMoody:
Just got to dive in and test things out locally. Everything looks awesome. I may also take and try to extend this pattern to Monsters, I think it'd work great there too!

I'll plan to leave this open for the next 12 hours or so in case anyone else has any comments, but after that I'll merge away on both! Don't worry about swapping them out now, I can plan to do that along with some other tweaks I'm planning to the nav. 

I'd also say we're fine to skip on pagination for now: it's a relatively small and defined datasource that doesn't seem to be causing any issues even on mobile at this point.

Thanks a bunch, let me know if you have any other ideas that we should be adding to the issues board.

eepMoody:
Sorry this took a bit to sort out. My local merge tool didn't like this. Took me an embarrassingly long time to realize it was because it was just a couple whitespace changes and it was ignoring them.

Merging in a moment as soon as the checks pass.

PR #294, opened 2019-03-31
Missing lines in 'Experience Points by Challenge Rating'

snan:
I added in the missing lines into
legacy-source-content/monsters/challenge-and-xp.rst

It says legacy, are you changing over the backend? Where is the new Single Point of Truth for this data? Been out of the loop b/c I hate discord


eepMoody:
Hey @snan, this point I'm very nearly done converting this whole thing over to django and using the files in data/WOTC_5e_SRD_v5.1 as a resource. BUT I will def port this over and make sure it gets into the new version as soon as it's done, because I believe it is missing there too. :D

snan:
Thank you so much♥

I don't have or like Discord so I'm kinda out of the loop.

PR #379, opened 2023-03-25
Fix linting

dprothero:
This fixes all the linting issues in the repo and then adds Husky, which will install the pre-commit hook whenever you run `npm install` so more people will have it locally.

Next, in a future PR, we'll run the linting during a GitHub Actions workflow so it gets checked even if someone bypasses the pre-commit hook.

Note: I use Prettier, so it did some other re-formatting beyond just fixing the linting issues. Let me know if this isn't preferable.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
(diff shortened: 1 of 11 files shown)


dprothero:
Closing as it has fallen behind. I am re-applying the work on the latest `staging` branch.

PR #381, opened 2023-05-16
fix: #377 blockquote styles

### LINKED ISSUE #377: Sage Advice Blockquotes are not rendered differently.
Originally: https://github.com/open5e/open5e-api/issues/51
image

Sage advice blocks are not visually different than the text around it.

This is actually a bug in the markdown file here:
https://github.com/open5e/open5e/blob/master/static/markdown/classes/rogue.md
Or possibly in the vue here:
https://github.com/open5e/open5e/blob/master/pages/classes/rogue.vue
The issue is that vue renderer does not seem to respect the blockquote.

peterbinks:
# Summary

Adding in blockquote styles to fix #377

I just made up some basic styles, but if there is an example I can update it to whats currently in use.

### Screenshots

Before:
image

After:
Screenshot 2023-05-15 at 10 42 35 PM
Screenshot 2023-05-15 at 10 42 30 PM

### Tech Notes

Styling elements directly obviously isn't the best approach, but it seemed the most straightforward, for now. This will effect all blockquotes on the site, which is what I think we'd want?

### DIFF
--- a/assets/_docs.scss
+++ b/assets/_docs.scss
@@ -101,3 +101,11 @@ h2.filter-header {
   justify-content: space-between;
   max-width: none;
 }
+
+blockquote {
+  background-color: #eaeaea;
+  border-left: .75rem solid #999;
+  padding: .1rem 1rem 1rem 1rem;
+  margin-top: 1rem;
+  max-width: 40rem;
+}


eepMoody:
This looks great!

PR #383, opened 2023-05-16
[WIP] Convert to Nuxt/Vue 3

eepMoody:
This one is a bit of a beast. I think it will be more productive to evaluate it based on running it vs a code review.


eepMoody:
As of current it runs, but the store is non-functional and needs to be converted over to `pinia`. I've had a few false starts, and I'm about ready to throw it at GPT4 and see if it can figure it out.

peterbinks:
@eepMoody looks like you deleted your .gitignore and all of node_modules are in there

eepMoody:
This branch is f'd due to a gitignore issue, remaking this pr as #384

PR #385, opened 2023-05-19
Nuxt 3 with Pinia

matthias-noetzli:
My attempt at getting the project to run on Nuxt 3 with Pinia, starting from a fresh Nuxt 3 project. It wasn't overly complicated but I had to touch a lot of files. A couple of things to note/that I wasn't sure about:

1. The store could probably be even simpler if we removed the getters, since all they do is directly return state. I kept them in for now because it meant less changes.
2. The LOAD_RACES and LOAD_SECTIONS actions in the vuex store had checks like `if (!context.sections)`. What was the intent there? I took them out for now, because I wasn't sure how to migrate them without knowing their purpose.
3. I took out vue-scrollto for now because I ran into this issue with Nuxt 3: https://github.com/rigor789/vue-scrollto/issues/736
4. I have not yet added the google analytics module. I did however see that it is no longer maintained.


eepMoody:
This is looking great! Just ran a test locally, and everything seems to be working perfectly as expected. Going to merge it into staging and keep an eye on it, make sure nothing unexpected crops up with the deployment.

I'm going to add on one small change to the homepage, just so it's easier to tell with 100% certainty that it's been deployed.

PR #396, opened 2023-05-24
Accessibility Improvements

matthias-noetzli:
- Added the lang attribute to the html element
- Fixed the sidebar list markup (nested lists need to be in an li-element).
- Added alt texts to the images used on the home page (patreon, discord, etc.)
- Fixed some style issues that had crept because nuxt 3 does not have an element with the __layout id.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div class="layout">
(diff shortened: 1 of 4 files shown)


eepMoody (approved):
Very nice! I will make an effort to find other accessibility opportunities as we expand the site, too.

PR #398, opened 2023-05-30
Accessibility improvements for the monster list, spell list and magic item list pages

matthias-noetzli:
- Added additional information for screen readers.
- The placeholder for the filter-input now stays visible while typing.
- Added a sorting indicator to the table headers.
- Changed sortable table headers to include buttons to make them keyboard accessible.
- Changed the magic item list to show only one list of results while being filtered.
- Adjusted placeholder and source tag colors to meet the minimum contrast requirements.
- Changed the heading on those pages to also be a h1 element and removed the h1 from the "Open5e "in the top left. There should always be only one h1 element per page and it should be different between pages, so the h1 containing "Open5e" wasn't ideal.

Some of my sources for the changes:
([https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/](url),  [https://www.scottohara.me/blog/2022/02/05/dynamic-results.html](url))

### DIFF
--- a/assets/_variables.scss
+++ b/assets/_variables.scss
@@ -25,5 +25,6 @@ $color-mana: #166c9c;
 $color-fog: #f4f4f4;
 $color-smoke: #d4d4d4;
 $color-granite: #888;
+$color-gray: #767676;
 $color-basalt: #333;
 $color-darkness: #111;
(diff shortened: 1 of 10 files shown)


eepMoody (approved):
Thank you for these! Great catches and much-needed improvements.

PR #407, opened 2023-06-20
Fixes #384 - Added slash to API URL template string 

### LINKED ISSUE #384: UI layer is making a 301 to prepend a / to API calls.
image

This is introducing a small delay on all API calls.

calumbell:
Fixes #384 

The problem wasn't with the prepending the slash, but that a slash was missing before the query param. I added the slash to the template string containing the endpoint in the search component. This appears to have fixed the problem.

### DIFF
--- a/pages/search/index.vue
+++ b/pages/search/index.vue
@@ -150,7 +150,7 @@ export default {
       this.loading = true;
       return axios
         .get(
-          `${this.$nuxt.$config.public.apiUrl}/search?text=${this.$route.query.text}`
+          `${this.$nuxt.$config.public.apiUrl}/search/?text=${this.$route.query.text}`
         ) //you will need to enable CORS to make this work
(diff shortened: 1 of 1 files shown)


eepMoody (approved):
Nice find! I was poking around and not finding this for ages!

calumbell:
Ah, looks like my fork was out of sync, and this issue has already been fixed in PR #400. Sorry about that, should have double checked!

eepMoody:
No worries, I hadn't noticed that either!

If you're looking to jump in, I could really use another set of sharp frontend-eyes on this PR:https://github.com/open5e/open5e/pull/404

It adds a bunch of new stuff related to our API calls, and I'm low-confidence that I got it all right!

calumbell:
I have some time this week, would be happy to take a look!

Issue #276, opened 2018-10-20
Make section/markdown pages more automatic


eepMoody:
The current markdown pages under Characters, Combat, Gameplay Mechanics, etc. are obviously super manual. It'd be nice if they were made into a more re-usable component or handled in an completely data-driven way.

calumbell:
I have some time this week and would be happy to take a look into this issue!

I am assuming that what we are after is something closer to how the `/classes` or `/races` route are currently structured. With a nested `index.vue` component containing links to all pages in that section and a `[id].vue` component for displaying each page? Let me know if I am on the right track

calumbell:
With #411 merged into staging I think we are good to mark this issue as closed!

PR #411, opened 2023-06-29
Feature/276 auto section md pages

calumbell:
This PRs fixes issue #276. The following subroutes were formly hard coded, and have been replaced with dynamic routes driven by data from the API

- /characters
- /classes
- /combat
- /equipment
- /gameplay-mechanics

And the following routes pull in data that already existed on the API, but wasn't being used on the front end:

- /feats
- /backgrounds

Several of these routes were pulling their data from markdown stored locally in the /public directory. These routes now pull their data from the API. The 12,000 or so removed lines was me removing this markdown where it was duplicated on the API. The only markdown files that were not duplicated on the API where the pages under the /combat route. I added these missing fields to the API in [this pull request](https://github.com/open5e/open5e-api/pull/288), but in until this is merged these routes will not work. In the mean time, I have conditionally rendered Combat heading in the navigation menu so that it will only appear when it is able to pull data from the API

This refactor turned up several bugs that while technically outside of the scope of the issue I went ahead and fixed as I discovered them. I'll pop comment on any of these fixes if they seem obscure.

This PR is already pretty substantial, so even though there are more things that I would like to add some polish to, they would be better served in seperate PRs. This is why I have left styling to a minimum (happy to pick this up later!). Also  Tailwind has been installed on the staging branch since I created this branch, so things could have gotten quite messy!

### Following up

Here are some things I noticed while working on this PR that would be worth looking into, possibly opening some issues.

**Subclasses**. These used to be appended to the end of the relevant class page, they now exist on their subroutes within each class route that can be access via links on the parent class' page. They are currently access via a list of links at the bottom of the `/classes/[className]`. It works, but it isn't pretty, and could definitely do with a restyle.
Screenshot 2023-06-29 at 10 42 45

**Race List in navigation.** Now that we are pulling this data from the API, the list of races in the navigation bar is way too long. It would probably be a good idea to refactor this (perhaps only showing WOTC SRD races, or instead we could show the different document sources that contain races in the list, and hovering over them would reveal a list of races available form that source?)

Screenshot 2023-07-02 at 12 22 01


components/SourceTag.vue
```diff
@@ -37,6 +46,7 @@ export default {
   margin-right: 3px;
   text-transform: uppercase;
   position: relative;
+  height: min-content;
```
calumbell, line 49: This property means that the a source-tag can be the child of an element with `display: block;` (ie. a header) without changing its height to fill all the vertical space available

pages/gameplay-mechanics/ability-scores.vue
calumbell, line 1: Replaced these hard-coded routes with dynamic routes using `pages/gameplay-mechanics/[section].vue`

pages/gameplay-mechanics/index.vue
```diff
@@ -2,53 +2,30 @@
...
-          <nuxt-link tag="a" to="/gameplay-mechanics/rest"> Rest </nuxt-link>
-        </li>
-        <li>
-          <nuxt-link tag="a" to="/gameplay-mechanics/saving-throws">
-            Saving Throws
-          </nuxt-link>
-        </li>
-        <li>
-          <nuxt-link tag="a" to="/gameplay-mechanics/time"> Time </nuxt-link>
-        </li>
```
calumbell, line 46: These links are now generated dynamically using data from the API

pages/races/index.vue
```diff
@@ -3,40 +3,32 @@
...
-          <nuxt-link tag="a" to="/races/half-elf"> Half-Elf </nuxt-link>
-        </li>
-        <li>
-          <nuxt-link tag="a" to="/races/half-orc"> Half-Orc </nuxt-link>
-        </li>
-        <li>
-          <nuxt-link tag="a" to="/races/halfling"> Halfling </nuxt-link>
-        </li>
-        <li>
-          <nuxt-link tag="a" to="/races/human"> Human </nuxt-link>
```
calumbell, line 31: These links are now generated dynamically from data pulled from the API

pages/sections/[id].vue
Comment (calumbell) [inline pages/sections/[id].vue:1]: We now no longer require this route because each of the section parents has its own subroute. This should make implementing breadcrumbing a little simpler

public/markdown/characters/advancement.md
calumbell, line 1: These local markdown files are no longer needed. We pull this data from the API now

store/index.js
calumbell, line 1: I added a few getters to the Pinia store. They do not add any new data to the store, but they handle the filtering of the different sections returned from the `/sections` API route. It made sense to me to move it here to avoid unnecesary duplication of the logic across various Vue components

components/MdViewer.vue
calumbell, line 1: Added a type to the `src` prop to make the linter happy. Also removed some console.logs

pages/appendix/index.vue
calumbell, line 1: This page contained links to pages displayed the sections with the parent "Running the Game". As all of these pages had routes of the form `/running/[slug]`, it didn't make sense to to list them on the `/appendix` route.

calumbell (commented):
Hi, I added some annotations to better explain what is going on in this PR. Hoping that this makes the reviewers' lives easier! Let me know if anything that I missed which would benefit from some more context

eepMoody (approved):
Looks all good! I've added issue #415 to track adding source selection to these. 

If you get to it before me, the pattern I'm going with is that _lists_ (like the subclasses on a class page, or the items in the sidebar) are filtered based on source selections, but pages for individual items (eg. the subclass) are not.

The thinking being that if you follow a link it should always work, but if you're browsing then we want to respect your choices RE: which sources you want to use. Hopefully the two places where the source is listed will make it clear that what you're using is outside your current selection.

Issue #139, opened 2016-03-29
Suggestion: compact statblock for monsters.


GAZ082:
Hi there. Would be nice to have a compact statblock for monsters, like those present in the two column SRD document from Wizards. Handy to cop and paste in campaign documents.

maxrothman:
Good idea! Once we've standardized the monster pages with a custom Sphinx domain, we can probably do this pretty easily.

txtsd:
@GAZ082 Do you mean like this?
```
STR  4 (-3)
DEX 16 (+3)
CON 11 (+0)
INT  2 (-4)
WIS  8 (-1)
CHA  6 (-2)
```

GAZ082:
I dont recall how it was 3.5 years ago! But i believe i was refering to something like this that, by the way, already looks like:

![image](https://user-images.githubusercontent.com/100194/68078862-d83c7d80-fdbd-11e9-9c5f-66c4e67784c8.png)

PR #412, opened 2023-06-30
Added page for compact monster stat layouts.

bruno-71:
Added link to compact page from monster page.  Also added .action-block to main.scss for reusability.

This is in response to issue #139.   It seems to be an old request, and I'm not sure what exactly the desired end result it.  But this is a start.  I just stripped out some of the formatting and decreased some spacing.

### DIFF
--- a/assets/main.scss
+++ b/assets/main.scss
@@ -4,11 +4,11 @@
 
 a {
   text-decoration: none;
-  @apply text-blue-700 hover:underline visited:text-purple-600;
+  @apply text-blue-700 visited:text-purple-600 hover:underline;
 }
 
 .side-note {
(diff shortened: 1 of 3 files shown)


eepMoody:
Very cool! I think we may later want to extract this out into a component and provide some sort of "compact / comfy" toggle, but this is an awesome foundation for that!

eepMoody (approved):

bruno-71:
Great...thanks!
By the way, I got the notifications that the "PR Validation" failed.  I'm still a little new to open source.  Do I need to fix something for that?  I seems like it was just complaining about some missing brackets.

eepMoody:
Nope, nothing to worry about there! Some more opinionated style rules got added while you were working on this. Just needed to to reorder a couple things, and I took care of it. No way you could have prepared for it. :)

PR #429, opened 2023-10-21
Monster Filters

markrmessmore:
Added a filter area for users to filter monsters by type, size, HP, challenge rating and name. If adequate, will explore creating similar filters for other desired pages.


eepMoody:
So this one looks great, but also includes a new Prettier version that's causing a dependency conflict with Tailwind. Is there any way to back that part out without breaking it?

markrmessmore:
Replaced by #433

PR #436, opened 2023-11-06
Added more options, specifics to the pagination.

markrmessmore:
Is this what you were looking for in terms of the QoL/UX options?


markrmessmore:
Including these options within another PR.

Issue #450, opened 2024-02-04
Some compact stat blocks crashing


drewdoesdev:
# Issue

When visiting some compact monster stat blocks with legendary actions, the page crashes.

https://github.com/open5e/open5e/assets/9469597/f5bef132-cda8-4f4e-865f-b7152c4cdacf

## Steps to reproduce

1. Visit a stat block with no legendary action or reaction found in the monster data.
2. Select Compact

### Expected
- Compact stat block is presented

### Actual
- Page is blank

PR #452, opened 2024-02-06
[FIX][OPEN5E][ISSUE-450] Fix for blank page on compact monster stat blocks

drewdoesdev:
Resolves [issue 450: Some compact stat blocks crashing](https://github.com/open5e/open5e/issues/450);

## Description
- Upon investigation, the issue seemed to be that the template for compact stat blocks was calling the length property of reactions and legendary actions, which return null from the data when undefined.  To resolve this issue, optional chaining was added to these property to prevent crashing.

### Production Site
**Gnoll**
image

**Aboleth**
image

**Lich**
image

### Local Site
**Gnoll**
image

**Aboleth**
image

**Lich**
image

### DIFF
--- a/pages/monsters/compact/[id].vue
+++ b/pages/monsters/compact/[id].vue
@@ -109,7 +109,7 @@
         <b class="action-name">{{ action.name }}. </b>
         <md-viewer class="inline" :text="action.desc" />
       </p>
-      <p>Monster Reactions: {{ monster.reactions }}</p>
+
       <hr v-if="monster.reactions?.length > 0" />
       <b v-if="monster.reactions?.length > 0">Reactions</b>
       <p
--- a/pages/monsters/compact/[id].vue
+++ b/pages/monsters/compact/[id].vue
@@ -89,6 +89,7 @@
         <challenge-render :challenge="monster.challenge_rating" />
       </div>
 
+      <!-- Special Abilities -->
       <hr v-if="monster.special_abilities?.length > 0" />
       <p
         v-for="ability in monster.special_abilities"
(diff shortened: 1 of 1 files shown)


eepMoody:
@drewdoesdev I'm trying to get our tests to re-run on this one.
Would you mind pushing a blank commit or something to try and get those to kick off? I can't figure out what happened here.

drewdoesdev:
Thanks you @calumbell and @eepMoody !  I've re-pushed my branch with some comments added above the stat block sections on the stat blocks to trigger a rebuild.  Let me know if anything ells is needed!

calumbell:
This is a sidebar, and very much outside the scope of the issue that this PR resolves, but I think we should take a second look at how we handle the compact statblocks.

The way thing stand, we will have to maintain two seperate monster pages for the regular and compact versions of the statblocks, and the only major difference between the two should be styling. This could be a problem long-term. My instinct is that there should only be one page with a toggle for switching between regular/compact displays of the statblocks, perhaps using a query param so that the compact block can be linked to directly.

The current URL structure for the compact block is also unhelpful for navigating the site ([/monsters/goblin](https://open5e.com/monsters/goblin) & [/monsters/compact/goblin](https://open5e.com/monsters/compact/goblin)), and breaks the Breadcrumbing by implying the existance of a `/monsters/compact` route.

I'll probably break this out into its own issue.

Thanks again to @drewdoesdev for making this fix!

drewdoesdev:
@calumbell No problem, though it seems like the build steps aren't running for some reason...trying to determine what's different between my MR and, say, this one which does have the steps https://github.com/open5e/open5e/pull/455

calumbell:
Fairly certain this is an issue on our end, looks like something has gotten stuck in our checks. I'll look into it when I am not very sleepy. Thanks again!

calumbell:
Alright, so still digging around for a solution to whatever is going on with our PR validation. But I have seen several reports that closing & re-opening the PR sometimes helps. So I am going to give that a go now

PR #455, opened 2024-02-13
Set custom titles for all pages

### LINKED ISSUE #454: Bugfix: page titles are always "Open5e"
On the [live site](https://open5e.com/) no matter which page you are visiting, the window/tab title will always be 'Open5e'.

It would be nice if the window title changed depending on the page being currently visited. Especially for people who make use of tab browsing and bookmarks.

For example:

- The title for the [fireball](https://open5e.com/spells/fireball) page could be **Fireball - Open5e** or  **Fireball - Spells - Open5e**
- The title for the [goblin](https://open5e.com/monsters/goblin) page should be something like **Goblin - Open5e**
 or **Goblin - Monsters - Open5e**
- The homepage should keep the title **Open5e**

The Nuxt [useHead()](https://nuxt.com/docs/getting-started/seo-meta#usehead) composable is proabably a good place to start looking

Sturlen:
Closes #454. Uses the same logic as the breadcrumbs, only in reverse. Works for all pages.
image

### DIFF
--- a/components/BreadcrumbLinks.vue
+++ b/components/BreadcrumbLinks.vue
@@ -75,4 +75,19 @@ const crumbs = computed(() => {
     })
     .filter((breadcrumb) => breadcrumb);
 });
+
+const BASE_TITLE = 'Open5e';
+
+const title = computed(() => {
+  if (crumbs.value.length === 0) {
+    return BASE_TITLE;
+  }
+  return (
+    crumbs.value
+      .map((crumb) => crumb.title)
+      .toReversed()
+      .join(' - ') + ` - ${BASE_TITLE}`
+  );
+});
+useHead({ title: title });
 </script>
--- a/components/BreadcrumbLinks.vue
+++ b/components/BreadcrumbLinks.vue
@@ -40,54 +40,5 @@
 </template>
 
 <script setup>
-import { useRoute } from 'nuxt/app';
-import { computed } from 'vue';
-const crumbs = computed(() => {
-  let url = '';
-  return useRoute()
-    .path.split('/')
-    .map((segment) => {
-      // ignore initial & trailing slashes
-      if (segment === '' || segment === '/') {
-        return;
-      }
-
-      // rebuild link urls segment by segment
-      url += `/${segment}`;
-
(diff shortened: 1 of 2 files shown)


calumbell (changes requested):
Thanks for making this fix, it works perfectly on my local branch!

My only misgiving is that, while it makes sense that you set the title in the `Breadcrumbs` component as you were reusing the logic there, the Breadcrumbs component isn't really the right place to be setting the page title. It makes much more sense to do this on the layout or on the specific page component.

The DRY-est solution I see is to refactor the logic used to generate the Breadcrumbs into its own [composible](https://vuejs.org/guide/reusability/composables) and use it in multiple places.

Something like `useBreadcrumbs()` or `useUrlSections()` that would take a URL string as a parameter and return an array of strings, each corrosponding to a breadcrumb. Perhaps with an `options` object where you could specify whether you wanted the array to be returned reversed, &c. 

Reckon this is something you'd be up for? Thanks again for the contribution!

Sturlen:
I've moved the logic for breadcrumbs and title up into the layout itself. I don't see a need to create a more complicated API for it unless there is a specific use case. Any child component/page can at any time overwrite the title with it's own call to `useHead()`.

calumbell (approved):
Brilliant, thanks for making these changes! Everything appears to be working on local.

I see your point about not wanting to needlessly overcomplicate the API, but composibles are a helpful tool for avoiding repetition while satisfying the single-responsibility principal, and are no more complex than refactoring logic into its own function. I am flagging this here because the layout component is already long and difficult to read, so we should be mindful when we move additional logic into it.

But that is a conversation for another day, and way outside of the scope for this PR. The titles look great. Especially happy that they can be overwriten at the page level. Thanks again for your contribution! 

(Total aside, but I have never used the `provide() / inject()` API. Looks super useful for avoiding prop drilling, but are there any advantages to using it over passing the crumbs as props to the `<breadcrumb-links>` component?)

calumbell:
Looks like the PR validation is stuck. I am going to close and re-open the PR to see if that fixes the issue

PR #456, opened 2024-02-14
Bugfix: source filtering for top-level pages

### LINKED ISSUE #415: Add source filtering to new class, race, combat, etc. endpoint calls
@calumbell introduced a number of awesome new dynamic sections in https://github.com/open5e/open5e/pull/411

These were developed in parallel with the new source-selection system, so they need to be updated to use the pinia store for requesting their resources, and should respect the source selection for the top-level pages (but not for individual items)

calumbell:
This PR closes #415 by making changes to certain top-level pages so that they respect the sources selected by the user. All pages now appear to be working as intended.

Aside: the Pinia store in its current state isn't easy to work with. Is this something that we would consider refactoring? Or perhaps even replacing with some of Nuxt's [built in features for caching](https://nuxt.com/docs/getting-started/data-fetching#caching-and-refetching), so long as we can get it playing nicely the other website features (source selection, &c)? I'd be happy to take the lead on looking into this.


calumbell:
Fixing merge conflict has broken PR validation. Closing and reopening this issue to retrigger tests

calumbell:
Closing this PR as the issue was fixed in #503

PR #458, opened 2024-02-17
Cypress e2e testing

### LINKED ISSUE #451: Feature: Unit Testing
The current version of the Open5e website doesn't feature any unit tests. Implementing these will help make the site more robust, and save many potential future headaches.

We do not have a testing library installed, so this will be a two-step process:

1) Select an appropriate testing library ([Jest](https://jestjs.io/) and [Vue Testing Library](https://testing-library.com/docs/vue-testing-library/intro/) have been suggested as places to start looking)
2) Write unit tests for each page to make sure that;
  2a) Data can be imported from the API
  2b) The page renders correctly

Sturlen:
Fixes #451 . Plan is to run e2e test that make sure new commits don't break the site. Still WIP.


Sturlen:
Cypress didn't work out as expected. Looking into other alternatives instead.

PR #462, opened 2024-02-20
Fix site generation errors due to Array.toReversed

Sturlen:
In my previous PR I used `Array.toReversed`, which turns out is not supported in Node 16. This creates issues when trying to generate the site using GH actions, which be seen here: https://github.com/open5e/open5e/actions/runs/7891045830/job/21534539552#step:6:186

This uses `Array.reverse` on a copy of the array instead.


calumbell (approved):
Looks good! I might even prefer this markup. Easier to read and less nesting

PR #463, opened 2024-02-21
Stop search page crashing when there is no search parameter.

Sturlen:
When `npm generate` is generating the search page, it does not add a search query. this causes the PR validation GH action to fail. I though I fixed this in PR #462, but it turns out the breadcrumbs function and the search result ordering both relied on the search query being there.

This is just a quick fix to let the PR validation function again, but the handling of search queries and breadcrumbs could do with more work at a later date.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -138,9 +138,10 @@ export default {
           const [title, queryParams] = segment.split('?');
 
           // extract & format the search params if on the /search route
-          const searchParam =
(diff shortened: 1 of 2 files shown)


eepMoody (approved):

calumbell (approved):
Thanks for getting this fixed! This bug has been giving me a reall headache. The PR works perfectly on my local machine, `npm run generate` now executes without any errors

PR #465, opened 2024-02-23
Monster saves only display if monster has proficiency

christophermiles:
Saving throws are now only listed when the monster has proficiency in the save (which the API signals by having `number|null` on the `[ability]_save` property

Aboleth saving throws on D&D Beyond:

Screenshot 2024-02-24 at 9 55 46 am

Updated saving throws on Open5E:

Screenshot 2024-02-24 at 9 58 01 am

Screenshot 2024-02-24 at 9 55 59 am

### DIFF
--- a/pages/monsters/[id].vue
+++ b/pages/monsters/[id].vue
@@ -291,7 +291,6 @@ const abilites = [
   shortName: ability.slice(0, 3),
   score: monster[ability],
   modifier: formatMod(calcMod(monster[ability])),
-  // save: monster[`${ability}_save`] ?? calcMod(monster[ability]),
   save: monster[`${ability}_save`],
 }));
 </script>
--- a/pages/monsters/[id].vue
+++ b/pages/monsters/[id].vue
@@ -47,7 +47,7 @@
     <!-- ABILITY SCORES -->
     <section class="max-w-96">
       <ul class="flex items-center gap-4 text-center">
-        <li v-for="ability in abilites" :key="ability.name">
+        <li v-for="ability in abilities" :key="ability.name">
           <span class="block font-bold uppercase">{{ ability.shortName }}</span>
           <span>{{ `${ability.score} (${ability.modifier})` }} </span>
         </li>
@@ -59,18 +59,20 @@
     <!-- SAVING THROWS AND ATTRIBUTES-->
     <section>
       <ul>
-        <li>
+        <li v-if="abilities.filter((a) => a.save).length">
           <span class="font-bold">Saving Throws </span>
+
           <span
-            v-for="ability in abilites.filter((a) => a.save)"
+            v-for="ability in abilities.filter((a) => a.save)"
             :key="ability.name"
+            class="after:content-[','] last:after:content-[]"
           >
             <span class="capitalize before:content-['_']">
               {{ ability.shortName }}
             </span>
             <span class="before:content-['_']">
-              {{ formatMod(ability.save) }} </span
-            >.
+              {{ formatMod(ability.save) }}
+            </span>
           </span>
         </li>
 
@@ -80,10 +82,10 @@
             v-for="(skill, key, index) in monster.skills"
             v-show="key !== 'hover'"
             :key="index"
+            class="capitalize before:content-['_'] after:content-[','] last:after:content-[]"
           >
-            {{ key.charAt(0).toUpperCase() + key.slice(1) }}
-            <span v-if="skill >= 0">+</span>{{ skill }}.
-            <span v-if="index < monster.skills.length - 1">, </span>
+            {{ key }}
+            <span v-if="skill >= 0">+</span>{{ skill }}
           </span>
         </li>
 
@@ -126,7 +128,11 @@
         class="action-block"
       >
         <span class="font-bold after:content-['.']">{{ ability.name }}</span>
-        <md-viewer inline="true" :text="ability.desc" />
+        <md-viewer
+          :inline="true"
+          :text="ability.desc"
+          class="before:content-['_']"
+        />
       </p>
     </section>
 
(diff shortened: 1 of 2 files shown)


calumbell (changes requested):
This is looking great, exactly what we were looking for. Cheers for picking this up so quickly!

Just a couple minor changes before it is perfect:

1) If a monster has no saving throw proficiencies (ie. `monster/goblin`) it would be good if we completely hide the **Saving Throws** label. Probably should have thought of that when opening the issue, mae culpa. Can probably be solved using a `v-if` directive against the length of the saving throws array.

screenshot from Open5e of the goblin statblock with an empty Saving Throws section

2) The saves are seperated by commas on the compact statblock, but period on the expanded one. The commas are what published statblocks generally use. Reckon we could replace those periods with commas? This can probably be done using Tailwind classes, something like `after:content-['.']` and `[&:not(:last-child)]:after:content-[',_']` iirc

Also, thanks for flagging the weird differences between the expanded and compact statblocks. I am planning to open an issue to collapse these into a single page with a toggle to change the style between expanded & compact. Might be a good one to pick up if you were looking for another project?

Thanks again for your contribution!

christophermiles:
Ah yep, makes total sense to hide the saving throw field completely if there are no saves!

I wondered why periods were used instead of commas. I assumed it was a stylistic decision as it seems to be used for other fields as well, and didn’t want to overstep the mark. But I think commas makes more sense.

Is there a reason why these punctuation marks are added as pseudo-content via CSS and not just output into the HTML?

christophermiles:
Also I think the idea of a compact-mode toggle is a good one, and not just to consolidate the code; the UX currently makes it hard to go back to the full stat block from the compact one

calumbell:
> Is there a reason why these punctuation marks are added as pseudo-content via CSS and not just output into the HTML?

Personal preference really.

I find that using it keeps the markdown neater and easier to read than, for example, conditionally rendered commas in span tags between each item. Functionally, the commas are styling, so it make sense to me to handle it with CSS. Just my two cents, if you have another solution and the markup looks good, then don't let me stop you!

christophermiles:
> > Is there a reason why these punctuation marks are added as pseudo-content via CSS and not just output into the HTML?
> 
> Personal preference really.
> 
> I find that using it keeps the markdown neater and easier to read than, for example, conditionally rendered commas in span tags between each item. Functionally, the commas are styling, so it make sense to me to handle it with CSS. Just my two cents, if you have another solution and the markup looks good, then don't let me stop you!

Nah that makes sense. There were other places where periods were used as delimiters; I’ll change them too for consistency.

pages/monsters/[id].vue
```diff
@@ -47,7 +47,7 @@
     <!-- ABILITY SCORES -->
     <section class="max-w-96">
       <ul class="flex items-center gap-4 text-center">
-        <li v-for="ability in abilites" :key="ability.name">
```
Comment (christophermiles) [inline pages/monsters/[id].vue:50]: Fixed typo in this property name

pages/monsters/[id].vue
```diff
@@ -59,18 +59,20 @@
     <!-- SAVING THROWS AND ATTRIBUTES-->
     <section>
       <ul>
-        <li>
+        <li v-if="abilities.filter((a) => a.save).length">
```
Comment (christophermiles) [inline pages/monsters/[id].vue:62]: Added filter

pages/monsters/[id].vue
```diff
@@ -59,18 +59,20 @@
...
           <span class="font-bold">Saving Throws </span>
+
           <span
-            v-for="ability in abilites.filter((a) => a.save)"
+            v-for="ability in abilities.filter((a) => a.save)"
             :key="ability.name"
+            class="after:content-[','] last:after:content-[]"
```
Comment (christophermiles) [inline pages/monsters/[id].vue:68]: Used Tailwind's `last` function to remove comma on last item

pages/monsters/[id].vue
```diff
@@ -80,10 +82,10 @@
             v-for="(skill, key, index) in monster.skills"
             v-show="key !== 'hover'"
             :key="index"
+            class="capitalize before:content-['_'] after:content-[','] last:after:content-[]"
           >
-            {{ key.charAt(0).toUpperCase() + key.slice(1) }}
-            <span v-if="skill >= 0">+</span>{{ skill }}.
-            <span v-if="index < monster.skills.length - 1">, </span>
+            {{ key }}
+            <span v-if="skill >= 0">+</span>{{ skill }}
```
Comment (christophermiles) [inline pages/monsters/[id].vue:88]: Used same approach as for saving throws to simplify the skills loop

pages/monsters/[id].vue
```diff
@@ -126,7 +128,11 @@
         class="action-block"
       >
         <span class="font-bold after:content-['.']">{{ ability.name }}</span>
-        <md-viewer inline="true" :text="ability.desc" />
+        <md-viewer
+          :inline="true"
+          :text="ability.desc"
+          class="before:content-['_']"
```
Comment (christophermiles) [inline pages/monsters/[id].vue:134]: Noticed there was a spacing issue here

pages/monsters/[id].vue
```diff
@@ -126,7 +128,11 @@
         class="action-block"
       >
         <span class="font-bold after:content-['.']">{{ ability.name }}</span>
-        <md-viewer inline="true" :text="ability.desc" />
+        <md-viewer
+          :inline="true"
```
Comment (christophermiles) [inline pages/monsters/[id].vue:132]: IDE highlighted an issue with the prop type here

pages/monsters/compact/[id].vue
```diff
@@ -48,14 +48,15 @@
 
       <hr />
 
-      <div v-if="getSaves">
+      <div v-if="getSaves.length">
```
Comment (christophermiles) [inline pages/monsters/compact/[id].vue:51]: Filtering for compact view

pages/monsters/compact/[id].vue
```diff
@@ -161,26 +162,22 @@ export default {
...
-          val: saveValue,
-          type: 'bonus',
+      ]
+        .filter((save) => this.monster[save.name + '_save'])
+        .map((val) => {
+          return {
+            name: val.display,
+            val: this.monster[val.name + '_save'],
+            type: 'bonus',
+          };
```
Comment (christophermiles) [inline pages/monsters/compact/[id].vue:179]: Used `map` here to simplify the code

calumbell (approved):
Looks good! Everything works as expected on my local machine. I'll go ahead and get this approved and merged. Thanks again for your contribution!

Issue #466, opened 2024-02-24
Refactor: expanded and compact monster stat-blocks


calumbell:
As of PR #412 we have had compact monster statblocks on the Open5e site. While this is a great feature, there are a few problems with its current implementation

- The expanded and compact statblocks are on completely different routes, and use different page components. This is proving difficult to maintain as the logic on the two pages has already begun to diverge.
- The URLs on the compact statblocks kind of torpedo the site breadcrumbing, implying the existance of a non-existant `/monsters/compact` route.

### Solution

As both the expanded and compact statblocks pull from the same API endpoint, I feel that these two pages (`/monsters/[id]` and `/monsters/compact/[id]`) should be collapsed into just the `/monster/[id]` page. The user should be able to toggle between the expanded and compact form of the statblock using a button. Ideally, the styles would match exactly the current expanded and compact styles.

Toggling this setting should only modify the page styles, and perhaps control the conditional rendering of elements that might not be appropriate to include in the compact stat-block. Additionally, it would be good if the toggle state was controlled by a query param (ie. `monsters/goblin?compact`) so that the compact statblock can be linked to directly.

### Notes

- To avoid link rot, a redirect from the `/monsters/compact/[id]` to the `/monsters/[id]` should be added.

christophermiles:
Keen to take a crack at this one!

christophermiles:
I wonder if it makes sense for this to be a global preference, like light/dark mode? What is the typical use case for viewing a compact stat block?

calumbell:
Awesome! Thanks for picking this one up

> I wonder if it makes sense for this to be a global preference, like light/dark mode? What is the typical use case for viewing a compact stat block?

I was thinking the same thing, but it would be worth considering who this would interact with the query parameter. 
The use case provided in the origin feature request  (#139) was that they are useful for copying and pasting directly into campaign documents without taking up too much space

PR #468, opened 2024-02-25
466 refactor expanded and compact monster stat blocks

christophermiles:
This PR addresses #466 by introducing a new toggle button at the top of the stat block that enables toggling between  'regular' and 'compact' display modes.

Previously this was achieved using two distinct components, and each component was assembling and styling the UI in different ways.

The original `/monsters/[id]/compact` route will now redirect to `/monsters/[id]?mode=compact`, which sets a toggle-able reactive value that is used to conditionally render parts of the template, and to modify the styling.

Changing the toggle will update the URL, meaning that a user can link to the compact mode version of a stat block.

The compact mode essentially bumps the text, line-height and margins down a smidge, and removes images. 

I'm not altogether sure the current compact mode fulfils the use case in the original ticket #139 as it's not all that easy to copy and paste, but this PR is at least a state toward unifying the logic for the two views.

I have placed the toggle button at top right for now, and tried to follow existing button styling (such as the filter button).

This PR also includes some changes to standardise how the 'inline list' elements (saving throws, skills) are spaced and delimited, as there were a variety of approaches used throughout the template.


pages/monsters/[id].vue
Comment (christophermiles) [inline pages/monsters/[id].vue:1]: I started adding a type for the `Monster` entity based on what I could glean from the model in the API repo, but the linter was complaining about TypeScript syntax in Vue files on commit

pages/monsters/[id].vue
```diff
@@ -1,12 +1,36 @@
 <template>
-  <main v-if="monster" class="docs-container container">
+  <main v-if="monster" class="docs-container container" :data-mode="mode">
     <!-- TITLE -->
-    <h1>{{ monster.name }}</h1>
-    <img v-if="monster.img_main" :src="monster.img_main" class="img-main" />
+    <div class="flex items-end justify-between gap-8">
```
Comment (christophermiles) [inline pages/monsters/[id].vue:4]: Added new button in the 'header' area to toggle the new `mode` value

pages/monsters/[id].vue
```diff
@@ -1,12 +1,36 @@
...
+      :src="monster.img_main"
+      class="img-main"
+    />
     <p class="italic">
       <span>
         {{
-          `${monster.size} ${monster.type} ${monster.subtype}, ${monster.alignment}`
+          `${monster.size} ${monster.type}${
+            monster.subtype ? ` (${monster.subtype})` : ''
+          }, ${monster.alignment}`
```
Comment (christophermiles) [inline pages/monsters/[id].vue:33]: Took the liberty of putting the subtype in brackets to follow the WOTC treatment

pages/monsters/[id].vue
```diff
@@ -19,24 +43,30 @@
     <section>
       <ul>
         <li>
```
Comment (christophermiles) [inline pages/monsters/[id].vue:45]: Removed some unnecessary spans here

pages/monsters/[id].vue
```diff
@@ -275,8 +295,12 @@ const monster = await useFetchArticle({
 const calcMod = (score) => Math.floor((score - 10) / 2);
 const formatMod = (mod) => (mod >= 0 ? '+' + mod.toString() : mod.toString());
 
+function uppercaseFirstLetter(string) {
+  return string.charAt(0).toUpperCase() + string.slice(1);
+}
+
```
Comment (christophermiles) [inline pages/monsters/[id].vue:301]: Was going to move this into `/utils` but that folder seems to contains a bunch of Python utilities, so I wasn't sure if that folder was meant to be used for JS utility functions

pages/monsters/[id].vue
```diff
@@ -290,11 +314,36 @@ const abilites = [
...
+  navigateTo({
+    path: `/monsters/${route.params.id}`,
+    query:
+      mode.value === 'compact'
+        ? {
+            mode: 'compact',
+          }
+        : null,
+  });
+}
```
Comment (christophermiles) [inline pages/monsters/[id].vue:340]: At one point I just had this function navigate to the new route, without the switch block, but because the component was not being re-mounted, the `mode` wasn't being updated from the query param, so I stuck with setting the `mode` ref and then doing the navigate just so there was a copy-able URL for the compact mode

pages/monsters/[id].vue
```diff
@@ -290,11 +314,36 @@ const abilites = [
...
 </script>
 
 <style scoped lang="scss">
 .img-main {
   float: right;
+  margin-block: 1rem;
```
Comment (christophermiles) [inline pages/monsters/[id].vue:346]: Just added some vertical margin on the floated image (incidentally I noticed that the floated image gets a bit wonky at certain screen widths...)

pages/monsters/[id].vue
```diff
@@ -305,4 +354,21 @@ const abilites = [
...
+  font-size: 1.44rem;
+}
+
+[data-mode='compact'] h2 {
+  font-size: 1.2rem;
+}
+
+[data-mode='compact'] hr {
+  margin-block: 0.5rem;
+}
```
Comment (christophermiles) [inline pages/monsters/[id].vue:373]: I based the smaller font-sizes on the default type scale at https://typescale.com and I reduced the vertical margins on the <hr>s

calumbell (changes requested):
Thanks for your contribution! The changes look great in the source code, but when I spin this PR up on local I am hitting a 500 error when I visit any `/monsters/[id]` route.

Here is what I see in my browser (Firefox 123.0) and terminal.

Screenshot 2024-02-28 at 08 48 50

Screenshot 2024-02-28 at 08 37 27

It looks like the error is occuring in the `formatMod()` function when it tries to run `.toString()` on a null value. Is this something that you could look into? I'd be happy to take a poke around, but I wouldn't want to step on your toes. You have worked on this component more recently than me, so probably have more insight into how everything works together.

By commenting out the calls to `formatMod()` on lines 102 and 116 of `monsters/[id].vue` I was able to get the page to render (sans saving throw mods). The expanded/contractor feature itself is looking great, exactly what we were looking for! Only have some very minor styling suggestions, (reducing margin above h2's and making the h1 small on compact are about it tbh) but this might be better to handle in its own issue.

Let me know if I can provide any extra info about this bug, and thanks again for getting this feature added!

christophermiles:
No problem, happy to fix that bug and implement those styling changes! May
be a day or so before I get to it. Thanks for the review and the testing!

christophermiles:
@calumbell I think the issue is that this code doesn't have the code that filters out saving throws with null values; I thought I had merged `staging` into this branch but I'll check again.

calumbell (approved):
Cheers for making those changes. Everything runs perfectly on local and it looks great!

And with that I think we are good to merge to staging, thanks for your contribution!

PR #477, opened 2024-03-08
fix: Monster skills not showing

### LINKED ISSUE #475: Bug: Skills not rendering correctly on monster statblocks
*This bug is on the `staging` branch, not the live site*

Monster skill proficiencies which are being returned by the API are not being rendered on the frontend.

### Reproduction
1) Checkout the`staging` branch of Open5e 
2) Navigate to a monster statblock with skill proficiencies, ie. `/monsters/goblin`
3) In the section underneath the saving throws, notice that there are no skill proficiencies listed

Screenshot 2024-03-06 at 09 25 36

4) If you compare this to the [/monsters/goblin API endpoint](https://api.open5e.com/monsters/goblin/), you will see that the goblin has proficiency in *stealth*. This should be visible on the frontend

Screenshot 2024-03-06 at 09 26 31

The bug is very likely on the `/monsters/[id].vue` page component

Sturlen:
Fixes #475. `monster.skills` is an object, not an array. Checking for a `length` property will always return `false` and hide the skills. This could have been avoided with more strict use of Typescript.

### DIFF
--- a/pages/monsters/[id].vue
+++ b/pages/monsters/[id].vue
@@ -102,8 +102,7 @@
             {{ formatMod(ability.save) }}
           </span>
         </li>
-
-        <li v-if="monster.skills.length">
+        <li v-if="monster.skills">
           <span class="font-bold after:content-['_']">Skills</span>
 
           <span


calumbell (approved):
That fixes it, cheers! I'll go ahead and get this merged

PR #481, opened 2024-03-10
Fix link to API repo

dprothero:


calumbell (approved):
Good spot, thanks for fixing! I'll get this approved and pushed to staging

PR #482, opened 2024-03-11
Feature: Search Scoped to Sources

### LINKED ISSUE #408: Search endpoint should be scoped to selected sources
Want to explore a half-measure on this.

User feedback is that they expect this to honor their source selections. But If we hide it something silently because of a source selection it might be assumed it wasn't there.

I believe the correct resolution here is a hybrid solution, where we execute the search query against the entire dataset.
Then on the UI layer, we can hide the non-matching results behind a "view additional results from other sources" kind of button?

image

image

calumbell:
This PR closes #408 by scoping searching to the specific sources selected by the user. This was achieved by through filtering the results returned by the API on the frontend, nothing about the call to the API has been changed.

In the example below, you can see that the number of results returned from allowed sources is visible underneath the page title.

Screenshot 2024-03-11 at 10 45 54

However, if you run the same query and prune back the sources, you see the following. Notice the prompt at the bottom of the screen. Clicking this will reveal sources returned by the API from sources outside of the user's current selection.

Screenshot 2024-03-11 at 10 47 25

While this PR resolves the initial issue, it does surface several other feature that we might want to implement in the future:

- Filtering search results by type (ie. only show me spells or monsters, etc.)
- Implement filtering via URL params on the backend to reduce response times.
- Add additional return fields to the search API endpoint so that, for example we can display *spell level* or *monster type & size* (ie. large monstrosity) to the search preview cards

### DIFF
--- a/components/SearchPreview.vue
+++ b/components/SearchPreview.vue
@@ -1,0 +1,98 @@
+<template>
+  <!-- Monster summary includes mini statblock -->
+  <div v-if="result.route == 'monsters/'">
+    <p>
+      <nuxt-link
+        tag="a"
+        :params="{ id: result.slug }"
+        :to="`/${result.route}${result.slug}`"
+        class="font-bold"
+      >
+        {{ result.name }}
+      </nuxt-link>
+      <span>{{ ` CR ${result.challenge_rating} | ` }} </span>
+      <em>{{ `${result.hit_points}hp, AC ${result.armor_class}` }}</em>
+      <source-tag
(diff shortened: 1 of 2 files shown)


calumbell:
Added a new commit to fix some build errors that occured during PR validation. Will now close and re-open the PR to re-run validation

calumbell:
Merge conflicts are probably not possible to untangle as a lot has changed on the site since this branch was created. Will scavenge what I can and create a new PR

Issue #473, opened 2024-03-03
Bug: Select Sources menu is missing dark-mode styles 


calumbell:
### Description and Recreation

When you click on the *Select Sources* button on the navbar, the modal menu that appears has the same styles in light and dark mode. The current styles are suitable for light mode, but not for dark.

*Light Mode*
<img width="400" alt="Screenshot 2024-03-03 at 12 25 31" src="https://github.com/open5e/open5e/assets/47755775/e7c6a6ac-04b5-4f66-91b4-b6bf7b639a6d">

*Dark Mode*
<img width="400" alt="Screenshot 2024-03-03 at 12 25 19" src="https://github.com/open5e/open5e/assets/47755775/feac2d7c-98ce-417c-8535-830abc13dad0">

### Solution

The Tailwind classes of the `SourcesModal.vue` component should be updated to include dark-mode styles. This can be done using the `dark:` selector. The light mode styles should be left unchanged

benhodges65:
![Fix](https://github.com/open5e/open5e/assets/77591479/6ba9c973-fc94-4338-b014-c48f1b3c4a60)
Hey @calumbell is this the expected look? If so I can open a PR. Also this is my first time contributing so if I should be putting images for feedback somewhere else I can do that, thanks!

calumbell:
Hi @benhodges65, thanks for getting involved. This looks great, exactly what we are looking for! I'll keep an eye open for when the pull request appears.

Best places to put images and mock-ups for feedback would be on the issue or on the pull request. Linking your pull request to the issue that it resolves by including a hashtag followed by the issue number (ie. this issue is #473) also helps us a lot to keep things organised.

PR #492, opened 2024-04-09
Add dark mode to sources

benhodges65:
This PR is for issue #473. It adds dark mode styling to the dialog panel and text of the sources modal to ensure the style stays consistent if the user is in Dark mode. Image below of example. 
Fix


calumbell (approved):
Thanks for your contribution. Runs perfectly on local. Will go ahead and get this approved and merged

Issue #488, opened 2024-04-04
Standardize all Vue pages to use the Composition API and <script setup>


Sturlen:
Currently pages are split between using the old Options API and the newer Composition API. Several pages are using both at the same time. This makes it difficult to maintain and refactor, and does not play well with newer libraries.

My suggestion is to move everything to the newer <script setup>, since that is the new standard. I should be able to implement the changes myself.

More context on the benefits of the Composition API: https://vuejs.org/guide/extras/composition-api-faq.html

calumbell:
Great idea, this has been bothering me for ages.

@Sturlen let me know if you'd like to have a go at fixing this issue, I can assign it to you. I'd also be happy to look into it also but I wouldn't have time for a few weeks yet.

Worth noting that getting the Pinia store to play nice with the Compositions API on page components might require a little finessing.

Sturlen:
@calumbell I'd be happy to give it a go. Assign away.

calumbell:
Done! (Not after accidently assigning to myself first)

calumbell:
With #496 and #497 merged to staging, I think we can close this issue as resolved. Thanks for making the fix @Sturlen!

PR #496, opened 2024-04-20
Refactor pages to compositon API

Sturlen:
Part of #488. Purely a refactor. Site should behave the exact same way as before,

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -111,222 +111,182 @@
   </div>
 </template>
 
-<script>
-import { useMainStore } from '../store/index';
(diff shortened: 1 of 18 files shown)


calumbell (approved):
Looks great to me!

Run perfectly on local. I checked out each of the top-level pages and at least one dynamic route for each page component, all appears to be working as intended. Building for production (`npm run build`) runs without error.

I'll go ahead and get this approved and merged to `staging`. Thanks for your work! Will take a look at you other pending PR (#497) either later today or tomorrow

PR #498, opened 2024-04-23
Fix monster sorting by CR

Sturlen:
Challenge rating currently sorts alphabetically, rather than numerically. This fix uses the numeric challenge rating available from the API to correctly sort them. 

The Problem:
image


calumbell (approved):
Nice fix! Works perfectly on local. I'll get it approved and merged

PR #495, opened 2024-04-16
494/bugfix slugify search links

calumbell:
This PR closes #494 by updating the redirects middleware correctly hyphenating URLs where a `section`'s parent had a name which included a space (ie. 'gameplay-mechanics).

While working on this issue, a related problem was surfaced. Sections on the API might have the parent `Gameplay Mechanics` or `Rules`, with only the former being actually index on a top-level page. This has been fixed by adding the Rules sections to the Gameplay Mechanics sections

PR #499, opened 2024-05-01
Use API v2 for Search

Sturlen:
Functionality remains the same. Uses the v2 endpoint but only returns v1 schema results. Links to gamplay mechanics are broken, but this is an existing issue fixed in PR #495.


calumbell (approved):
After spinning this up on local and having a poke around, everything looks good! Functionality appears to be as before. Nice work, i'll go ahead and get this merged to `staging`

PR #503, opened 2024-06-05
Replace all data fetching with Vue Query

Sturlen:
This replaces the custom fetching and caching solution made with Pinia with Vue Query. A lot of the changes made were to move logic out .vue files and into dedicated composables. This should make fetching much easier to reason about and avoid a lot of edge cases in the future.

There should be no changes in user experience, other than filters fields now being able to set back to "any" without clearing the whole all filters. Not all pages have consistent way of handling loading states, but this was an existing problem and should be fixed separately.

Most changed lines are from the `package-lock.json` file moving around dependencies due to installing Vue Query and it's Nuxt plugin. Underscore was also removed due to it no longer being used anywhere.

I've tested all interactions I can think of, but I'd like more eyes on it to make sure I didn't miss anything.

Squashed commit history because it contains a lot of different iterations that are not very useful anymore. [original branch](https://github.com/open5e/open5e/compare/staging...Sturlen:open5e:query)


calumbell:
Amazing work, i'll see about getting this reviewed over the next few days. Thanks for your contribution!

But there is a problem. With Node 16 now being officially deprecated, it looks like our PR validation is borked. PR #491 should fix this issue, but is still awaiting review (I am the author of said PR, so can't review it. Anybody else with the appropriate permissions have a moment to take a look? 👀). Until then, we might have to put a pin in merging this to `staging`

calumbell:
With #491 approved and merged to staging, I am going to try closing and re-opening this PR to rerun validation.

Sturlen:
Feel free to push any changes you need.

calumbell:
Alright, took a couple of goes (had to fix some linting errors while I was at it) but it looks like we are now clearing PR validation.

Will take a stab at reviewing the code tomorrow, although it we might be blocked from merging as I have committed to the PR

calumbell (approved):
This looks great! It will make working with the API so much easier.

Everything is running as expected on local, nothing glaring jumps out at me when looking over the code.

PR #504, opened 2024-06-06
Feature: Dice Roller

### LINKED ISSUE #337: Add random number generation to simulate dice rolls
For example, I'm looking at the page for young brass dragons: https://open5e.com/monsters/young-brass-dragon

It would be nice if any text that describes dice rolls would be a hyperlink (bold and red would look nice) that when clicked perform and display the roll by generating a random number.

For example, the `STR 19 +4` when clicked would generate a random 1d20 + 4.
The `Perception +6` would generate a random 1d20 + 6.
The `+7 to hit` when clicked would generate a random 1d20 + 7.
The `2d6 + 4` when clicked would generate a random 2d6 + 4.

Probably you could automatically create a random number generating link for any any regex match of `\d+d\d+( \+ \d)?`.

calumbell:
This PR closes #337 by adding a dice roller to the Open5e website.

Screenshot 2024-06-06 at 14 13 38

The roller can be used in two different way:
1) It can be added into page/component templates by `useDiceRoller()` composable as an on-click event handler for a component. (For an example, look at the HP on the `monster/[id]` route)

2) It can be inserted into markdown strings by passing the `:use-roller="true"` parameter to the `<md-viewer>` component. This extends the markdown viewer cmpnt to match common dice signatures using regex and then wrapping them in the `<inline-roller>` component. (For an example, look at the actions on the `monster/[id]` route)

Currently the results are displayed using the `alert()` function, which isn't a great user experience. But this PR is already big enough, and we could always circle back later and consider a better design once we have had a bit of time to sit with the feature a little.


calumbell:
Closing and reopening to re-run PR validation following merge conflict resolution

eepMoody (approved):
Very cool, lgtm!

PR #510, opened 2024-06-14
Bugfix: removed unnecessary import statements for defineEmits & defineProps

calumbell:
This PR patches a couple of Vue compiler warnings that were popping up when running the test server.

screenshot of a terminal window running the nuxt dev server, there are two prominent compiler warnings

`defineProps` and `defineEmits` were being imported in a few components. Since refactoring to the Vue compositions API (PR #496) this is no longer necessary.

The fix was as simple as removing these import statements. The compiler errors are gone and the site appears to work exactly as it did before.


calumbell:
Closing and reopening issue to re-run validation for new commits

eepMoody (approved):

Issue #508, opened 2024-06-13
Feature: pagination on the /monsters top-level page


calumbell:
Loading times on the https://open5e.com/monsters route are currently extremely slow. On a solid internet connection (~200 Mbps) I am observing >4s load times. Taking a look at the debug console, and it looks like this is due to this route requesting every monster in when it loads. Assuming no filters are applied, that is close to 2900 entries in one go!

One solution to this would be to add pagination to this page – chunking the monster list into blocks of 25-50 and then only requesting them when needed. A recent update to our data fetching infrastructure (PR #503) will likely make this fix much, much simpler. 

By my reckoning, this fix can be split into three tasks:

1) Update data fetching composables (`useAPI()`, etc.) to support pagination. This might involve passing an optional parameter to specify which pages to return.
2) Update UI so that the user can navigate between pages (UI on `/pages/spells/index.vue`might be a good starting point)
3) Play around with various pagination values (page size, number of pages to load at a time, etc) until the user-experience is smooth.

Once this is fixed, we will likely roll out similar updates to the `/spells` and `/magic-items` routes.

NB. While pagination exists on the [spells top-level page](https://open5e.com/spells), this is purely a UI feature. The page still requests every spells in one go, which leads to long load times.

Sturlen:
With the new standard API table component in place, we can use that to implement pagination for all pages that use it. I think is best done through a new `usePaginated` composable inside the table component that handles all the logic. 

@calumbell If you haven't already started work on this, then I'd like to be assigned. Already have a working prototype so it should go pretty smooth.

calumbell:
@Sturlen done!

PR #515, opened 2024-06-19
Feature: standard tables components for API results

### LINKED ISSUE #509: Feature: Use standard tables for all large list pages
Currently the pages that list a large amount of items (monsters, spells and magic items) are all structured differently. Some are grouped by first letter, some by spell level and some in table. 

I suggest using the same table components and filtering system for all of these. This makes the user experience more consistent and simplifies development. Features like pagination #508 will then only have to be implemented once.

This can also easily be expanded to all list pages, but is less necessary for routes with only a couple items.

calumbell:
This PR closes #509 by introducing standardised components for displaying results returned on from the API on the top-level pages of Open5e.

Screenshot 2024-06-19 at 09 30 42

Previously, many of the top-level pages handled how they displayed results from the API in idomatic ways. This was identified as an obstacle in implementing a pagination system #508.  Now each page uses the `<api-results-table>` component to render this data.

### Code Example

Screenshot 2024-06-19 at 09 29 40

```
<api-results-table
  v-if="monsters && monsters.length > 0"
  :data="filtered_monsters"
  :cols="['type', 'cr', 'size', 'hit_points']"
  endpoint="monsters"
/>
```

JSON API data is passed to the table using the `data` prop. The `cols` prop  controls which fields from the `data` object to display on the table. As this code example was taken from the `/monsters/` route and these columns have been chosen to match the live site. Different pages uses different arrangements of columns.  If the `cols` prop is omitted, the table will only contain a single sortable column of the names of the API results.  The `endpoint` prop is only used to create the hyperlinks in the table.

The `<api-results-table>` depends on two sub-components; the existant `<sortable-table-header>` (these new tables are sortable in the same way as the old ones) and the new `<api-result-row>` (which formats the table rows, builds the links to specific API items, etc.)


calumbell:
Merge conflicts resolve. Closing and reopening issue to re-run validation

calumbell:
More merge conflicts resolved. Once more closing and reopening to re-run PR validation

eepMoody (approved):

Issue #517, opened 2024-06-21
Bug: dark mode styles on `<MonsterFilterBox>` and `<MagicItemFilterBox>`


calumbell:
### Description
The `<MonsterKilterBox>` and  `<MagicItemFilterBox>` do not look right on dark mode. The text colour changes from black to white, but none of the background colours of the form elements change. This results in unreadible text due to low colour contrast.

### Recreation

To recreate this bug, checkout the `staging` branch and visit either the `/monsters` or `/magic-items` routes. Click the *show filters* button. Try using it in light mode, then try using it in dark mode.

<img width="1620" alt="Screenshot 2024-06-21 at 08 55 18" src="https://github.com/open5e/open5e/assets/47755775/1ccf9b26-316d-419e-92cf-57b0719a40dc">

<img width="1618" alt="Screenshot 2024-06-21 at 08 55 28" src="https://github.com/open5e/open5e/assets/47755775/d0f9bfea-d119-4251-b041-32c0399237e3">

The forms in the first and second image contain the same data, but in dark mode you cannot read it because the white text matches the background colour of the input element.

### Solution
Add dark mode styles to the problematic form elements in the `<MonsterKilterBox>` and  `<MagicItemFilterBox>`. We have used the [TailwindCSS dark: selector](https://tailwindcss.com/docs/dark-mode) to handle this elsewhere on the website.

calumbell:
This issue was fixed when #518 was merged to staging

PR #518, opened 2024-06-21
fixed filter box text dark mode

pmclooneuf:
fixed Issue #517 by making the text for the filter box black while in dark mode

### DIFF
--- a/components/MagicItemFilterBox.vue
+++ b/components/MagicItemFilterBox.vue
@@ -1,6 +1,6 @@
 <template>
   <!-- FILTER BOX -->
-  <div class="filter-header-wrapper flex flex-wrap bg-smoke px-2">
+  <div class="filter-header-wrapper flex flex-wrap bg-smoke px-2 dark:text-black">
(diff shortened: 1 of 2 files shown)


calumbell (changes requested):
Hi, thanks for your contribution. Being able to actually read the text in the filter box is a substantial improvement to the user experience!

Screenshot 2024-06-22 at 09 56 41

However, the element itself doesn't really match the rest of the page. Light text on a dark background is more in keeping with the look of the site in dark-mode. Would you be able to update the styles so that:
- Input fields are light text against a dark background (small colour as the website background?)
- The background for the entire filter menu is dark (but lighter than the site background)

Our theme colours are defined in the `tailwind.config.ts` file in the project route, might be helpful.

Thanks again for your contribution! lmk if you have any questions

eepMoody:
I'm gonna go ahead and merge this one in, then do a fast-follow today to get the styling more in line with our current standard! It's something I'd like to put some time in this week, and I don't want to create any conflicts with this work. :)

eepMoody (approved):

PR #525, opened 2024-06-27
Feature: search scoped to sources

### LINKED ISSUE #408: Search endpoint should be scoped to selected sources

calumbell:
This PR closes #408 by applying by filtering to search results based on the current sources that the user has selected. Results from other sources can be viewed by clicking a button at the bottom of the `/search` page.

The filtering by source currently takes place on the front end, in the script tag for `/pages/search/index.vue` page component.

### Demonstration 
With all sources selected you might see something like this:
Screenshot 2024-06-27 at 09 34 16

Here is what the search page looks like if your search is returning only a few results due to your sources (NB. the highlighted snippet returned by the API is only rendered if the search term is not included in the article's title)

Screenshot 2024-06-27 at 09 35 05

Clicking on the view results from other sources button exposes a second list:
Screenshot 2024-06-27 at 09 35 14

### Closing Thoughts
There are a few vector to futher optimisation to the `/search` page that spring to mind. These were not pursued in this PR as they would have required making changes to our data fetching infrastructure, putting them somewhat out of the scope of this PR.

- Currently, in scope and out of scope results are all fetched on page load and filtered on the front end. It would be more efficient to only fetch the in scope results on page load and only fetch the out of scope results when needed.
- Server-side pagination could speed up load times (see issue #508)

### DIFF
--- a/components/SearchResult.vue
+++ b/components/SearchResult.vue
@@ -3,22 +3,23 @@
 <template>
   <li class="py-2 text-base">
     <!-- Row title -->
-    <h3 class="mt-1 flex align-middle text-lg">
+    <h3 class="mt-1 flex items-center align-middle text-xl">
       <nuxt-link
         tag="a"
         :to="`${getRoute(result.object_model)}/${result.object_pk}`"
       >
         {{ result.object_name }}
       </nuxt-link>
+      <source-tag :text="result.document.key" :title="result.document.name" />
       <span
-        class="font-sans text-sm uppercase text-granite before:mx-2 before:text-lg before:content-['|']"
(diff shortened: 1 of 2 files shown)


eepMoody (approved):
Sorry this one took a few, one of the more complicated ones recently, but it looks great!

PR #530, opened 2024-07-12
Pagination for top level list pages

### LINKED ISSUE #508: Feature: pagination on the /monsters top-level page

Sturlen:
Closes #508. This PR makes all list pages paginated, to avoid the long load times when fetching thousands of items for a page. Filtering and sorting pages will now all fetch new data from the server have also been switched to fetch only the required data, rather than being client side. All fetching logic is now handled by the <api-results-table> itself, to simplify and make data fetching consistent across all pages. The spell list for each class are unchanged as they are a separate issue.

Will need to run `npm install` as I've installed the `@vueuse/core` for it's debouncedRef function, which greatly simplifies input denouncing. Could look into just extracting the functionality we need but it would take some time.

This PR is a draft right now because debouncing the filter input, to avoid making a new request for each keypress, is still in progress. I've run into some issues with Vue reactivity and the debouncing not playing well together and am still debugging. There's a also some logging that needs to be cleaned up.

Let me know if you have any feedback or want any changes done.


Sturlen:
Waiting for https://github.com/open5e/open5e-api/pull/483 before it's ready to be pushed to beta.

Sturlen:
Had to remove tests that mocked internal page API usage, due to these being replaced with new ones. Proper tests for these pages should be done in a later PR.

Other tests that test page rendering are passing.

pages/races/index.vue
```diff
@@ -1,25 +1,16 @@
...
-      </div>
+    <div class="filter-header-wrapper">
+      <h1 class="filter-header">Races</h1>
     </div>
-    <api-results-table :data="races" endpoint="races" />
+    <api-results-table
+      endpoint="races"
+      :api-endpoint="API_ENDPOINTS.sections"
+      :cols="['document__title', 'document__slug']"
+    />
```
eepMoody, line 10: Currently this is returning a list of sections instead of a list of races. Does it need a filter like the other `sections` related items have?

pages/running/index.vue
```diff
@@ -1,11 +1,20 @@
...
+    />
   </section>
 </template>
 
 <script setup>
 import ApiResultsTable from '~/components/ApiResultsTable.vue';
-const { data: sections } = useSections('Rules');
+
+const filter = ref({
+  parent__in: 'Rules',
```
eepMoody, line 18: This also seems to be returning a list of all sections instead of just the ones from the sidebar list

eepMoody (changes requested):
This looks awesome overall, and I think it's going to make the v2 switch _sooooo_ much easier to manage! Couple pages seem to be not filtering to the right content, but otherwise this looks gtg to me!

Would like to see a merge from staging before we do a final approval, since there's new test actions that would run against it.

eepMoody:
Sorry, IDE confusion with the auto-created issue 😵‍💫

Sturlen:
Thanks for catching those oversights. Both have been corrected. Staging has been merged in and the AI code review seems to have run, though I don't see any result from it.

calumbell:
@Sturlen This looks great, amazing work! Seems to work more-or-less as expected on local, but I did notice a couple to problems during snagging:

1) Results on the `/conditions` page doesn't appear to respect source selection
2) When no results are returned on a page, the page number indicator reads *Page 1 of 0*. My instinct would be to hide this bit of text when no results are returned, or change this to *Page 0 of 0*

Screenshot 2024-08-07 at 17 48 08

calumbell (approved):
Thanks for making those changes, everything looks good to me. Happy to go ahead an approve this.

This is going to improve load times on our top-level pages so, brilliant work!

calumbell:
@Sturlen Just a few final notes. Really little ones, I promise!

1) Merge conflict is probably due to bumping the Nuxt version last week. Merging from staging and running `npm i` should fix it.

2) You left a few `console.log()` debug statements in your code (`components/ApiResultsTable.vue`, `components/ApiTableNav.vue`, `composables/api.ts`, `/pages/magic-items/index.vue`, `pages/monsters/index.vue`)

calumbell:
Looks like PR validation is stuck. I'll close and reopen the PR to rerun it

PR #531, opened 2024-07-15
Feature: Unit Testing

### LINKED ISSUE #451: Feature: Unit Testing

calumbell:
Closes #451 by introducing rudimentry unit testing using the `nuxt-testing-utils` ([link](https://nuxt.com/docs/getting-started/testing)) package. Under the hood, this uses Vitest to run unit tests. This package also supports end-to-end testing should we wish to add that in the future.

The tests can be run using the command `npm run test`:

Screenshot 2024-07-15 at 20 09 54

(I have experimented with adding the test command to the Husky pre-commit config. I have commented this out for now as the tests do take quite a long time to run and might benefit from refinement before enabling this feature on `staging`)

At this point the tests are simple sanity tests to make sure that nothing is exploding in the page components:

1 - Check that each page can mount successfully
2a - For pages that display a single Open5e item, test that mocked data is rendered correctly in HTML.
2b - For pages that display multiple Open5e items (ie. top-level pages), test that they render the correct number of hyper-links given an input array of mocked Open5e data.

## Next Steps

These tests are only a start, and there is plenty of work to do to expand their scope and robustness.
a) Unit testing for Components
b) Unit testing for Composables
c) End-to-end testing

There is one part of this test suite that requires further work that I felt fell outside of the scope of this PR. Testing on the `/search` route was not possible due to errors being thrown whenever the `/pages/search/index.vue` page component was mounted:

Screenshot 2024-07-12 at 23 05 39

I had a good go at trying to fix this problem, but no dice. I left my work on these tests in situ with the problematic lines commented out. Perhaps somebody else might have better luck figuring out this bug.


.husky/pre-commit
```diff
@@ -15,4 +15,7 @@ if ! git diff-index --quiet HEAD --; then
   echo "Lint fixes applied and staged."
 else
   echo "No lint fixes needed."
-fi
\ No newline at end of file
+fi
+
+# Run tests
+# npm run test
```
calumbell, line 21: Un-commenting this line will run tests during pre-commit. I have left this commented out because the tests were quite slow

tests/unit/pages/search.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+ * I am leaving my work on these unfinished tests in this PR. The code that
+ * triggers this bug is commented out below. It might be easier to resolve once we
+ * have implemented unit testing for the `SearchResult` component.
+ */
+
+// const page = await mountSuspended(SearchPage);
+
+// test('/search page can mount', async () => {
+//   expect(page).toBeTruthy();
+// });
```
calumbell, line 27: This test was the problematic one mentioned in the PR itself. If this doesn't get fixed before being merged to `staging` i'll create a new issue for it

eepMoody (approved):

PR #542, opened 2024-08-03
Update search endpoint and formatting

eepMoody:
This update adds a trailing slash to the search endpoint definition, to avoid a 301 error (identified by @augustjohnson) with the new v2 endpoints

It also cleans up the display of search previews by:
- switching to a yellow highlight instead of a gray one
- making the snippet dark enough to pass accessibility checks
- removing markdown table formatting (pipes and 3+ dashes in a row) from preview snippets
- rendering the snippets as markdown (to include bold or other useful formatting)
  - this also requires stripping newlines from the snippet, which has been done
  

Before:
  
image

image

   
After:

image

image

### DIFF
--- a/components/SearchResult.vue
+++ b/components/SearchResult.vue
@@ -10,12 +10,10 @@
       >
         {{ result.object_name }}
       </nuxt-link>
-      <source-tag :text="result.document.key" :title="result.document.name" />
-      <span
-        class="font-sans text-sm uppercase text-granite before:mx-2 before:content-['|']"
-      >
+      <span class="ml-2 font-sans text-sm uppercase text-granite">
         {{ result.object_model.match(/[A-Z][a-z]+/g).join(' ') }}
       </span>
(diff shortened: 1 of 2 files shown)


components/SearchResult.vue
```diff
@@ -10,12 +10,10 @@
...
         {{ result.object_name }}
       </nuxt-link>
-      <source-tag :text="result.document.key" :title="result.document.name" />
-      <span
-        class="font-sans text-sm uppercase text-granite before:mx-2 before:content-['|']"
-      >
+      <span class="ml-2 font-sans text-sm uppercase text-granite">
         {{ result.object_model.match(/[A-Z][a-z]+/g).join(' ') }}
       </span>
+      <source-tag :text="result.document.key" :title="result.document.name" />
```
eepMoody, line 16: Re-ordered so elements go from large & most-important (type) to small & least-important (source tag)

Issue #539, opened 2024-07-30
Bug: trailing comma in monsters without alignment


calumbell:
When viewing a monster that  has an alignment, it is displayed prominantly right underneath the monster's name and inline with its type, size, &c, preceded by a comma.

`/monsters/goblin` route on `staging`:
<img width="925" alt="Screenshot 2024-07-30 at 08 08 49" src="https://github.com/user-attachments/assets/c95abe3d-5775-4166-b015-8afeeafdde0d">

However, many monsters in the Open5e database do not have an alignment (ie. all of monsters from LUA 5e). In these cases the website render

<img width="925" alt="Screenshot 2024-07-30 at 08 08 36" src="https://github.com/user-attachments/assets/ba3b5e74-498b-4049-ba50-01c6413f5286">

This trailing comma looks goofy, and should only be rendered when an alignment is returned by the API. There are a couple different ways to handle this, but in my opinion using some kind of Tailwind class (something like `before:content-[',_']`) is the tidiest way to proceed.

PR #543, opened 2024-08-03
539 Bug: trailing comma in monsters without alignment

eepMoody:
https://github.com/open5e/open5e/issues/539

Without alignment: 
image

With alignment:
image

### DIFF
--- a/pages/monsters/[id].vue
+++ b/pages/monsters/[id].vue
@@ -31,7 +31,7 @@
         {{
           `${monster.size} ${monster.type}${
             monster.subtype ? ` (${monster.subtype})` : ''
-          } ${monster.alignment ? `, ${monster.alignment}` : ''}`
+          }${monster.alignment ? `, ${monster.alignment}` : ''}`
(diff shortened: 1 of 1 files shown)


calumbell (dismissed):
Hey thanks for getting this submitted! Works a treat on local, but I am concerned that the markup for this quite difficult to read with the template strings and ternaries. Would it be possible to refactor this is a slightly clearer way?

One approach might be to split it up into a series of conditionally rendered spans and use Tailwind to add in the punctuation. Perhaps something like the following would be a place to start:
```
<span>{{ `${monster.size} ${monster.type}`}}</span>
<span
  v-if="monster.subtype"
  class="before:content-['_('] after:content-[')']"
>
  {{ monster.subtype }}
</span>
<span v-if="monster.alignment" class="before:content-[',_']">
  {{ monster.alignment }}
</span> 
```

calumbell (approved):
Looks great! Thanks for making those changes

PR #555, opened 2024-08-18
Bugfix: disable SRI on dynamic routes

calumbell:
This PR should act as a work around for a bug which many user have reported where navigating to a dynamic route (ie. `/classes/bard` or `/monsters/goblin`) throws an SRI error. As this bug is likely due to a weird interaction with Cloudflare it is not really possible to thoroughly test this PR on local, but the test server and build process appear to work without issue.

NB. this PR does not fix the bug, but it should mean that the website is still usable while we investigate solutions for this error.


calumbell:
Closing and reopening to rerun PR validation

PR #590, opened 2024-09-26
Refactor: remove relict files from project root

calumbell:
This PR goes some of the way to counteract bloat in the Open5e root directory due to new features which require new configuration files (vitest, tailwind, etc.) by removing files that are no longer in use on the website. The goal here is to reduce cognitive load on any programmer who is new to the repo.

The following files have been removed from the directory root, being artefacts from when the back-end was also handled in this directory:
- `create_certs.py`
- `docker-compose.yml`
- `user-data.conf`

Additionally the following files were move to the `/public` directory with the other static files
- `logo.png` (moved to `public/img/logo.png`)


eepMoody (approved):

eepMoody:
This is looking good to me! Feel free to merge away!

PR #604, opened 2024-11-12
540 styles dark mode for error page

christophermiles:
I abstracted the `useHead` code that was applying a `dark` or `light` value to the body class list, so it would apply those values in the `error.vue` page as well (since this doesn't use the `default` layout`)

I also restructured the HTML in the `error.vue` slightly to make the page layout feel a little closer to the main page layout.

And added Tailwind classes to `error.vue` to apply the dark theme as it has been applied elsewhere.

Screenshot 2024-11-12 at 4 56 51 PM

Screenshot 2024-11-12 at 4 57 00 PM

Screenshot 2024-11-12 at 4 57 18 PM

Screenshot 2024-11-12 at 4 57 24 PM


.gitignore
christophermiles, line 1: Just added this to prevent JetBrains files being tracked

composables/useThemeHead.ts
christophermiles, line 1: I got a warning saying the `children` is deprecated and that `textContent` or `innerHtml` is the preferred alternative but haven't changed as part of this work

Screenshot 2024-11-12 at 5 18 05 PM

composables/useThemeHead.ts
```diff
@@ -0,0 +1,19 @@
+/* load theme preference before page and insert them into the head. This is to
+ * avoid flicker on load. 1st check local storage, then check media preferences
+ */
+
+export function useThemeHead() {
+  useHead({
+    script: [
+      {
+        children: `
```
christophermiles, line 9: I got a warning saying the `children` is deprecated and that `textContent` or `innerHtml` is the preferred alternative but haven't changed as part of this work

Screenshot 2024-11-12 at 5 18 05 PM

error.vue
```diff
@@ -1,76 +1,85 @@
 <template>
-  <div class="container">
-    <h1>Critical Failure!</h1>
-    <h2>Error {{ error?.statusCode }}</h2>
-    <p class="font-italics">{{ error?.message }}</p>
+  <div class="min-h-dvh text-darkness dark:bg-darkness dark:text-white">
+    <div class="container w-full sm:max-w-[40rem] sm:pl-56">
+      <div class="p-4">
```
christophermiles, line 4: I changed the structure here slightly so that the error page content is positioned roughly how the main site content is positioned

error.vue
```diff
@@ -1,76 +1,85 @@
...
+        <h1>Critical Failure!</h1>
 
-    <button class="font-bold text-red hover:text-blood" @click="handleError">
-      Return to Homepage
-    </button>
+        <h2>Error {{ error?.statusCode }}</h2>
 
-    <p>Try searching for what you need in the site menu</p>
+        <p v-if="error?.message" class="font-italics" v-text="error.message" />
```
christophermiles, line 9: Not rendering this `<p>` at all if there's nothing to render

error.vue
```diff
@@ -1,76 +1,85 @@
...
-        xml:space="preserve"
-        x="0px"
-        y="0px"
-        class="die-roll"
-      >
-        <g>
-          <path
-            d="M201.4,285.94c-29.92-12.17-59.86-24.31-89.76-36.56c-3.46-1.42-5.05-0.76-6.77,2.76
+        <button
+          class="font-bold text-red hover:text-blood dark:text-indigo-200 dark:hover:text-red"
```
christophermiles, line 12: Added some dark mode classes on the link based on classes used elsewhere

error.vue
```diff
@@ -82,6 +91,9 @@ const props = defineProps({
     type: Object || undefined,
   },
 });
+
+useThemeHead();
```
christophermiles, line 95: Set `dark` or `light` on the body class list using the existing code

error.vue
```diff
@@ -82,6 +91,9 @@ const props = defineProps({
     type: Object || undefined,
   },
 });
```
christophermiles, line 93: Would be kind of nice to use a type for the error; I am assuming this would be an instance of `NuxtError`, which extends `H3Error`:

```ts
declare class H3Error<DataT = unknown> extends Error {
    static __h3_error__: boolean;
    statusCode: number;
    fatal: boolean;
    unhandled: boolean;
    statusMessage?: string;
    data?: DataT;
    cause?: unknown;
    constructor(message: string, opts?: {
        cause?: unknown;
    });
    toJSON(): Pick<H3Error<DataT>, "data" | "statusCode" | "statusMessage" | "message">;
}
```

The `error.default` and `error.type` properties in the current props definition don't seem to be used in the template or JavaScript, but the `error.statusCode` and `error.message` properties _are_ used in the template, but not expected in the props!

I wonder if we can just do:

```
<script setup lang="ts">
  defineProps<{ error: NuxtError }>()
// ...
```

calumbell (approved):
Nice! Tests are passing and it runs as expected on local Happy to approve this.

Refactoring the `useHead()` composable into `useThemeHead()` is a good idea. What are you feelings about refactoring it further so that the new composable can return getter and setter methods to interact with the theme `useTheme()`. Something like:

```
{ getTheme, setTheme } = useTheme();

// toggle theme
if (getTheme() === 'dark') setTheme('light');
else setTheme('dark');
```
What are your thoughts?

Good spot on the `children` prop on DOM Element being a deprecated, probably a good idea to update.

Typing the error object also seems like a good idea! It is a low-hanging fruit and will make further changes much easier

christophermiles:
> What are you feelings about refactoring it further so that the new composable can return getter and setter methods to interact with the theme `useTheme()`

I don't think there is a `useTheme()` currently? All the theme management stuff is happening in `<ThemeSwitcher>`; I didn't add `<ThemeSwitcher>` to the error page as I wasn't sure there was a need to be able to set the theme from that page.

But I couldn't find a `useTheme`. were you proposing changing `useThemeHead` to `useTheme` so that it both does that `useHead` stuff, and also exposes a getter and setter that `<ThemeSwitcher>` can interact with, instead of `<ThemeSwitcher>` doing the getting and setting?

christophermiles:
> Good spot on the children prop on DOM Element being a deprecated, probably a good idea to update.

Should this be a separate ticket, or worth doing as part of these changes?

christophermiles:
> Typing the error object also seems like a good idea! It is a low-hanging fruit and will make further changes much easier

@calumbell Similarly, would this be a separate ticket, or OK to do as part of this PR?

I don't seem to be able to import types or get proper type hinting in my IDE, is the only thing. I wasn't sure if there was an issue with the project ESLint, but I am getting ESLint errors everywhere that TypeScript is used.

calumbell:
> But, yeah, I couldn't find a `useTheme` so I'm a bit confused. Were you proposing changing `useThemeHead` to `useTheme` so that it both does that `useHead` stuff, and also exposes a getter and setter that `<ThemeSwitcher>` can interact with, instead of `<ThemeSwitcher>` doing the getting and setting?

My bad! `useTheme()` doesn't exist, I was speaking hypothetically. Like, it would be cool if something like this existed, and this is how you might use it. Sorry for the confusion, my post wasn't very clear. I'll open a new ticket where i'll explain my thoughts my fully.

> @calumbell Should this be a separate ticket, or worth doing as part of these changes?

If it is a case of change the name of the `script` field, then we may as well tackle it here. If it requires any more work than that it should probably go into a seperate issue.

> I don't seem to be able to import types or get proper type hinting in my IDE, is the only thing. I wasn't sure if there was an issue with the project ESLint, but I am getting ESLint errors everywhere that TypeScript is used.

I am pretty sure that ESLint isn't configured correctly, I also get buckets of TypeScript errors which I generally judiciously ignore (which entirely defeats the purpose of using TS). This is definitely something worth fixing. Might be an idea to open a new ticket (feel free to do this, or I can do it tomorrow)

--

@christophermiles Sorry if this sounds too much like nit-picking. Thanks again for your contribution!

christophermiles:
@calumbell All good! I'll have a look at those other suggestions and raise them as new tickets if I think it's warranted.

christophermiles:
Also, ESLint configs never seem to work properly for me either when I set them up in new projects 😆

But I might take a look to see if anything obvious jumps out

components/BreadcrumbLinks.vue
christophermiles, line 1: I moved some of the classes out of `BreadcrumbLinks`, `ThemeSwitcher` and `SidebarToggle` that were specific to how these are laid out on the main page, to make these components more reusable and less coupled to their parent layout

components/SidebarToggle.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <button
-    class="m-1 h-min rounded-full bg-fog px-3 py-2 hover:bg-smoke dark:bg-basalt hover:dark:bg-granite sm:hidden md:px-4 md:py-3"
+    class="flex h-10 w-10 items-center justify-center rounded-full bg-fog hover:bg-smoke dark:bg-basalt hover:dark:bg-granite"
   >
-    <Icon name="heroicons:bars-3" class="text-center" />
+    <Icon name="heroicons:bars-3" />
```
christophermiles, line 5: Fixed some existing issues with the vertical centring of the icon within the button

components/ThemeSwitcher.vue
```diff
@@ -1,39 +1,25 @@
...
+    :aria-roledescription="
+      theme === 'light' ? 'Dark Mode Toggle' : 'Light Mode Toggle'
+    "
+    @click="toggleTheme"
+  >
+    <Icon
+      :name="theme === 'light' ? 'heroicons:sun' : 'heroicons:moon'"
+      class="text-center text-lg"
+    />
+  </button>
```
christophermiles, line 18: Tidied this up to use a single button that has its properties and classes updated according to the current theme

components/ThemeSwitcher.vue
```diff
@@ -1,39 +1,25 @@
...
-  const body = document.documentElement;
-  if (body.classList.contains('dark')) {
-    body.classList.remove('dark');
-    localStorage.setItem('theme', 'light');
-  } else {
-    body.classList.add('dark');
-    localStorage.setItem('theme', 'dark');
-  }
-};
+const { theme, toggleTheme } = useThemeSwitcher();
```
christophermiles, line 24: Moved the theme management code into a composable in case we want to set the theme from other components

composables/useThemeSwitcher.ts
christophermiles, line 1: New composable for managing the theme

composables/useThemeSwitcher.ts
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  function getFallbackTheme(): Theme {
+    return window.matchMedia('(prefers-color-scheme: dark)').matches
+      ? 'dark'
+      : 'light';
+  }
+
+  function getTheme(): Theme {
+    if (!import.meta.client) {
```
christophermiles, line 17: Have to do this to avoid errors if we're not in browser-land yet

error.vue
```diff
@@ -84,15 +84,12 @@
...
-    type: Object || undefined,
-  },
-});
+<script setup lang="ts">
+import { NuxtError } from '#app';
 
-useThemeHead();
+defineProps({
+  error: NuxtError,
+});
```
christophermiles, line 93: I couldn't use `defineProps<>` or do `import type { NuxtError }` using the current ESLint setup

layouts/default.vue
```diff
@@ -58,7 +58,9 @@
         <!-- Navigation Links -->
         <ul class="text-inherit text-white">
           <li v-for="section in routes" :key="section.title">
-            <nav-link :to="section.route"> {{ section.title }} </nav-link>
+            <nav-link :to="section.route">
+              {{ section.title }}
+            </nav-link>
```
christophermiles, line 63: Auto-formatting

layouts/default.vue
```diff
@@ -92,10 +94,10 @@
...
         <!-- Site Header -->
 
-        <div class="flex h-12 align-middle">
-          <sidebar-toggle @click="toggleSidebar" />
-          <breadcrumb-links class="flex-grow" />
-          <theme-switcher class="inline-block" />
+        <div class="flex h-12 items-center gap-1 px-2 sm:pl-8">
+          <SidebarToggle class="sm:hidden" @click="toggleSidebar" />
+          <BreadcrumbLinks class="flex-grow" />
+          <ThemeSwitcher />
```
christophermiles, line 100: Layout of these elements is now defined where they are used here in this parent component instead of coupling the layout to the components themselves by having margin classes within the components

nuxt.config.ts
```diff
@@ -14,7 +14,7 @@ export default defineNuxtConfig({
         { charset: 'utf-8' },
         { name: 'viewport', content: 'width=device-width, initial-scale=1' },
         {
-          hid: 'description',
+          key: 'description',
```
christophermiles, line 17: This was suggested by my IDE; apparently `hid` is deprecated

nuxt.config.ts
```diff
@@ -33,6 +33,18 @@ export default defineNuxtConfig({
...
+          type: 'text/javascript',
+          innerHTML: `
+          if (localStorage.theme === "dark" || (!('theme' in localStorage) && window.matchMedia("(prefers-color-scheme: dark)").matches)) {
+            document.documentElement.classList.add('dark')
+          } else {
+            document.documentElement.classList.remove('dark')
+          }
+        `,
+        },
+      ],
```
christophermiles, line 47: Instead of having `useThemeHead` and using it in `default` layout and `error`, I'm just setting it here. This also improves the situation with the flicker before the theme kicks in.

christophermiles:
@calumbell Made some more updates based on your suggestion — please see comments above!

calumbell:
@christophermiles These changes look great! Exactly what I had in mind for the `useThemeSwitcher()` composable. Thanks for your efforts, go ahead and merge to `staging`!

christophermiles:
@calumbell Great to hear!

PR #607, opened 2024-11-12
529 ux retain filter settings when navigating back to monsters page

christophermiles:
This PR syncs the current magic items filter and monster filter with local storage (I implemented the magic item filter syncing because it was the only other page that was setting an 'initial filter').

I had to make some changes to `<ApiTableFilter>` to make the initialised filter values from local storage reflect in the UI, as previously the inputs were just tracking their own internal state.

@calumbell Wonder if I should set all the filterable pages up like monsters and magic items?

@calumbell I'm not sure how I feel from a UX perspective about the filters persisting across sessions; I feel like it's handy within a session, maybe when viewing monsters then returning to the list, but maybe annoying to have to clear the filters on each visit? (Especially since we don't have a UI for clearing all the filters!)

The only way around this would be a composable for each filterable page (since each filter has different fields) that uses `useFilterState` and shares its own state across pages

Alternatively, `useFilterState` could track separate filter state for all known filterable lists. We would then need to move those state objects outside of the composable function so that they become singleton state. But this couples `useFilterState` very tightly to the specific filters.


components/ApiTableFilter.vue
```diff
@@ -9,7 +9,6 @@
       <input
         :id="search?.name"
         :name="search?.name"
-        default=""
```
christophermiles, line 12: Not a valid attribute (?)

components/ApiTableFilter.vue
```diff
@@ -30,10 +29,10 @@
         :key="field.name"
         :name="field.name"
         class="cursor-pointer bg-transparent fill-red text-center"
-        value=""
+        :value="field.value"
```
christophermiles, line 32: Setting `:value` so that values set at initialisation (eg. from local storage) are reflected in UI

composables/useFilterState.ts
```diff
@@ -1,34 +1,53 @@
 import { debouncedRef } from '@vueuse/core';
 
+export type FilterStateOptions<T> = {
+  initialFilters?: T;
+  debounceTimeMs?: number;
+  localStorageKey?: string;
+};
```
christophermiles, line 7: Use an options object instead of params

composables/useFilterState.ts
```diff
@@ -1,34 +1,53 @@
...
       (value) => value !== undefined && value !== ''
     ).length;
   });
 
-  const canClearFilter = computed(() => enabeledFiltersCount.value > 0);
+  const canClearFilter = computed(() => enabledFiltersCount.value > 0);
 
   function clear() {
-    filter.value = { ...initialFilters };
+    setFilter({});
```
christophermiles, line 37: If we're initialising with what's on local storage, then clear should literally clear all filters (clear is not actually used anywhere currently)

composables/useFilterState.ts
```diff
@@ -1,34 +1,53 @@
...
   function update(key: string, value: any) {
-    filter.value = { ...filter.value, [key]: value };
+    setFilter({ ...filter.value, [key]: value });
   }
 
   return {
-    clear,
-    update,
-    enabeledFiltersCount,
-    filter: computed(() => filter.value),
```
christophermiles, line 31: Wasn't sure why this computed was needed when `filter` is already reactive?

composables/useFilterState.ts
```diff
@@ -1,34 +1,53 @@
...
+    clearFilter: clear,
+    updateFilter: update,
```
christophermiles, line 48: Renamed for clarity and scoping when used in template code

pages/classes/index.vue
```diff
@@ -13,7 +13,7 @@
       />
     </div>
     <api-table-filter
-      :update-filters="update"
+      :update-filters="updateFilter"
```
christophermiles, line 16: Renamed for more clarity (we might have another `update` function that does something else)

pages/magic-items/index.vue
```diff
@@ -13,15 +13,17 @@
...
       :search="{
         name: 'Search Magic Items',
         filterField: 'name__icontains',
+        value: filter.name__icontains,
       }"
       :select-fields="[
         {
           name: 'Rarity',
           filterField: 'rarity',
+          value: filter.rarity,
```
christophermiles, line 26: Pass value through so inputs reflect current state when initialised with initial values (eg. from local storage)

pages/monsters/[id].vue
```diff
@@ -38,7 +38,7 @@
         {{ monster.alignment }}
       </span>
 
-      <!-- 
+      <!--
```
Comment (christophermiles) [inline pages/monsters/[id].vue:41]: I think my IDE just auto-formatted this...

calumbell (changes requested):
Thanks for getting these changes made! I have checked out the PR on local and had a poke around the test server. Great work so far, but I noticed a few bugs. Would you be able to address these before merging?

1) Navigating to the `/magic-items` or `/monsters` route from other routes works fine, but if you navigate directly to these links OR you reload these pages you hit a 500 error. The error message reads that `localStorage is not defined`.

Screenshot 2024-11-13 at 11 18 49

(This isn't an issue on the `/spells` route so I am fairly certain that this was introduced in this PR)

2) The value for the *search by name* field doesn't appear to be retained  in the same way that other fields are when you navigate back to the top-level page after viewing a single-article page. From a UX perspective I think this makes sense.

---

Regarding your other point about UX, I agree with you that it makes sense for filter state should reset on a refresh. It feels like this is what it is doing currently, but due to the 500 error when reloading on the `/magic-items` and `/monsters` routes I cannot confirm this.

All your other changes seem clear and well reasoned. Thanks for taking the time to add comments the PR, makes my life much easier!

christophermiles:
@calumbell Ah, yep, can take a look at those!

components/ApiTableFilter.vue
```diff
@@ -9,8 +9,8 @@
       <input
         :id="search?.name"
         :name="search?.name"
-        default=""
         placeholder="Search..."
+        :value="search?.value"
```
christophermiles, line 13: Current search (eg. from local storage) now loaded into search input

components/ApiTableFilter.vue
```diff
@@ -69,7 +70,8 @@
   </div>
 </template>
 
-<script setup>
+<script setup lang="ts">
+// TODO: Set up types for these props once ESLint supports use of defineProps<> so we can pass in generics for the filters we're using
```
christophermiles, line 74: Added TODO note

components/BreadcrumbLinks.vue
christophermiles, line 1: These changes came from #604

components/SidebarToggle.vue
christophermiles, line 1: Changes from #604

components/ThemeSwitcher.vue
christophermiles, line 1: Changes from #604

composables/monsters.ts
```diff
@@ -14,10 +16,6 @@ export const DefaultMonsterFilter: Readonly<MonsterFilter> = {
   type: '',
 };
 
-export function copyDefaultMonsterFilter(): MonsterFilter {
-  return { ...DefaultMonsterFilter };
-}
-
```
christophermiles, line 20: This wasn't being used anywhere

composables/useFilterState.ts
```diff
@@ -1,34 +1,63 @@
...
+        JSON.stringify(filterToSet)
+      );
+    }
+  }
+
+  if (options?.initialFilters) {
+    setFilter(options.initialFilters);
+  } else if (
+    options?.localStorageKey &&
+    import.meta.client &&
```
christophermiles, line 30: Added check here to resolve 500 error on page load

composables/useThemeSwitcher.ts
christophermiles, line 1: Changes from #604

error.vue
christophermiles, line 1: Changes from #604

layouts/default.vue
christophermiles, line 1: Changes from #604

nuxt.config.ts
christophermiles, line 1: Changes from #604

pages/magic-items/index.vue
```diff
@@ -2,26 +2,28 @@
   <section class="docs-container container">
     <div class="flex justify-between">
       <h1 class="my-2 w-full">Magic Items</h1>
-      <api-table-nav
-        :page-number="pageNo"
-        :last-page-number="lastPageNo"
+      <ApiTableNav
+        :page-number="pageNo || 1"
+        :last-page-number="lastPageNo || 1"
```
christophermiles, line 7: Because of how the props are set up, I was getting ESLint errors here

pages/magic-items/index.vue
```diff
@@ -2,26 +2,28 @@
...
     </div>
 
-    <api-table-filter
-      :update-filters="update"
+    <ApiTableFilter
+      :update-filters="updateFilter"
       :search="{
         name: 'Search Magic Items',
         filterField: 'name__icontains',
+        value: filter.name__icontains,
```
christophermiles, line 20: Pass through the current value of the filter to persist it

christophermiles:
@calumbell I've pushed some changes to address the issues you raised, thanks for spotting those!

> Regarding your other point about UX, I agree with you that it makes sense for filter state should reset on a refresh. It feels like this is what it is doing currently

Because we're persisting using localStorage, the filter state currently does _not_ reset on a refresh for the monsters and magic items pages where we're now persisting the filter settings. So if you refresh, the filters will be initialised with the last used value.

We could either:

- add a 'clear' button
- only persist the filters until refresh

Adding a clear button would be relatively straightforward I think, and changing the behaviour to persist the filters until refresh would mean ditching the local storage, and tracking individual state for each kind of filter. I think this would be doable; I imagine that `useFilterState` could have a singleton `Map` ref, with filter objects keyed by filter type (like the local storage keys I'm currently using).

I'd probably lean towards the latter. What do you think would be the better UX?

christophermiles:
@calumbell No sure why but the tests seem to be stuck?

calumbell (changes requested):
@christophermiles Thanks for addressing those bugs. Checking this out on local and I am not hitting those 500 res codes, and the search by name field state is retained in the same way.

## New Bugs

However, I did notice a new bug. All other top-level pages that include any filtering logic apart from `monsters` and `/magic-items` are throwing 500 errors. The error message reads `Cannot read properties of undefined (reading 'defaultFilter')` (nice dark mode error page styles btw)

Screenshot 2024-11-14 at 10 46 20

Getting this error on the following routes:

- http://localhost:3000/classes
- http://localhost:3000/spells

## UX

Regarding UX considerations for resetting the form, I also prefer the second option over the first, but I think using both would be the best.

A 'Clear Filters' button should be simple to implement, as the `clearFilter()` method is already implemented on the `useFilterState()` composable. I don't think the UI wants to be especially visually intrusive, just a small cross in an accent colour somewhere in the `<api-table-filter>` component should suffice.

Persisting the filters for only certain navigation types (yes for backward and forward, no for navigate or reload) looks to be a little more involved. A singleton in `useFilterState()` could work, but we might require greater granularity to only clear the filters on certain navigation types.

Another solution might be to keep the filter state in local storage, but clear them on a hard-reload using conditional logic. There might be something in the [useRouter() composable](https://nuxt.com/docs/api/composables/use-router) or the [PerformanceNavigationTiming API](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming) to help determine precisely when a hard reload takes place. Once that is done it is just a case of wrapping the clearFilter() function in an if-statement.

Don't take my words as gospel, they are just suggestions, and unopinionted ones at that. Feel free to to figure out your own way to solve this issue!

calumbell:
@christophermiles Tests tend to get stuck if you make additional commits to a branch after the PR has been created. Usually closing and re-opening the PR will fix it...

christophermiles:
Thanks @calumbell — sorry for missing those bugs. I'll fix those and add a clear button and have a think about the persistence thing. My initial thought is maybe check the from route and only persist when navigating to a filterable top level route like monsters if coming from a monster/:id route.

christophermiles:
@calumbell I've added the clear filter button. It only appears when there is an active search or filter on a filterable page, and I modelled its appearance on the pagination controls.

I'm also persisting the filters across the session but not on page refresh. (This ended up being a simple matter of moving the filter ref in the composable outside the composable function; each filterable page therefore has its own instance of `useFilterState`.)

I also simplified my changes to `ApiTableFilter` by passing in the return value of `useFilterState` as a prop, so we can use its `canClearFilter`, `clearFilter`, `updateFilter` etc methods and properties. This seems to be working well and means it's simpler to use the composable and component together without a lot of boilerplate.

Let me know what you think!

christophermiles:
Actually, I might need to test something else

calumbell (changes requested):
Thanks for making those change. The UX on the filter state retention is exactly what I imagined, nice work!

I found a couple of bugs when testing out on local. Hopefully these should be pretty simple fixes. Once they are fixed we can merge!

1) The *clear filter* button looks great on dark mode, but in light mode it retains its dark mode styles. Reckon we could add in some light mode styles also?

2) I am hitting an other 500 error when I navigate to the `/equipment` page. The error message reads `Cannot read properties of undefined (reading 'debouncedFilter')`

Screenshot 2024-11-15 at 11 35 27

The 500 errors I previously encountered on the `/classes` and `/magic-items` routes are no longer there. Perhaps a similar fix?

Great work so far!

christophermiles:
That second error will be an easy fix (I must have missed updating that one when I did some refactoring; apologies).

I modelled the clear filter button on the pagination controls so I _think_ they're behaving the same as those with respect to dark / light mode, but I'll take another look. (Maybe we could abstract out that button component so it's easier to keep the styles in sync.)

christophermiles:
@calumbell Just pushed a fix for the 500 error

I realised with the button I was using the styles for an 'inactive' pagination control, my bad.

christophermiles:
@calumbell I was thinking of making an `IconButton` we could use for those pagination nav buttons, the clear filter button, and also the theme switcher and sidebar toggle buttons?

calumbell (approved):
Looks great! Works well on the test server. Tests pass, `build` and `generate` scripts run without error. UX is good. Thanks for your work and your forbearance in the face of all th changes requested.

I have approved this PR, we are good to merge!

christophermiles:
Thanks for all your input and testing efforts @calumbell! It's been a fun feature to work on!

PR #610, opened 2024-11-16
608 dx fix eslint config

### LINKED ISSUE #608: DX: Fix ESLint config
Currently ESLint reports errors with things like:
- importing types
- TypeScript in general ("parsing error, unexpected token)

Just doing some local testing, replacing:

```js
module.exports = {
  root: true,
  extends: [
    // https://github.com/vuejs/eslint-plugin-vue#priority-a-essential-error-prevention
    'plugin:vue/vue3-strongly-recommended',
    'plugin:vue/strongly-recommended',
    'plugin:vue/vue3-recommended',
    'plugin:vue/recommended',
    'prettier',
  ],
  // required to lint *.vue files
  plugins: ['vue', '@typescript-eslint'],
  // add your custom rules here
  rules: {
    eqeqeq: 'off',
    // curly: 'error',
    // sometimes you need to use double quotes. ie. string contains apostrophes
    // quotes: ['error', 'single'],
    'vue/multi-word-component-names': 'off',
    'vue/html-self-closing': [
      'error',
      {
        html: {
          normal: 'always',
          component: 'always',
          /* Vue advises that void tags omit the terminal slash, but this
           * conflicts with Prettier */
          // void: 'never',
          void: 'always',
        },
      },
    ],
  },
};
```

with:

```js
module.exports = {
  root: true,
  extends: ['@nuxt/eslint-config'],
};
```

Immediately resolved the issues with TypeScript producing ESLint errors.

christophermiles:
This PR adds the [official Nuxt ESLint module](https://eslint.nuxt.com/packages/module) in place of the previous manual ESLint configuration. This means we will now be using the more modern flat ESLint config.

#### Removal of Prettier

Prettier has been removed as part of this change. We were only using the single-quotes rule from Prettier, and a plugin for Tailwind formatting. 

The single quote rule can also be achieved using [https://eslint.style](https://eslint.style), which is built into the Nuxt ESLint module (see [https://eslint.nuxt.com/packages/module#eslint-stylistic](https://eslint.nuxt.com/packages/module#eslint-stylistic)).

The Tailwind plugin we were using was also available as `eslint-plugin-tailwindcss` so I have incorporated that into the ESLint config to achieve the same result.

#### Removal of SASS files

I have also transposed the SASS CSS to the main `tailwind.css` file.

#### Additions from recent PRs

The addition of the recent `useFilterState` PR that went into staging is not yet present in the target `feature/v2-api` branch so the changes from the `useFilterState` branch will appear in the change list of this PR

#### Formatting changes

The bulk of the file changes are a result of running `lintfix` over the files with the new ESLint config. Mostly these are the removal of semi-colons.

#### Other changes

- The linter suggested using `never` in place of `any` in some TypeScript definitions
- The Tailwind linter recommended combining some classes like `w-6 h-6` into `size-6`, and made other suggestions and fixes


.husky/pre-commit
```diff
@@ -1,9 +1,6 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-# Run pretty-quick to format staged files
-npx pretty-quick --staged
-
```
christophermiles, line 6: I removed this because we're no longer using Prettier, but wasn't sure how best to replace this

bin/generate.js
christophermiles, line 1: - Removed unused imports
- Lint fixes

composables/api.ts
christophermiles, line 1: `any` has been changed to `never` at the suggestion of the linter

styles/tailwind.css
christophermiles, line 1: I've moved the CSS declarations from the SASS files here

I think we could use the Tailwind prose plugin in place of a lot of the rules for setting spacing between headings.

I think there is probably more we could do to consolidate the CSS and leverage Tailwind, but that is probably outside the scope of this PR.

calumbell:
@christophermiles this looks great, nice work! It is also pretty big. It will take a bit of time to review, but I am on it

christophermiles:
@calumbell No worries, and no rush. I realise nobody actually asked for this but I thought it might be nice to update the ESLint and remove the SASS dependency

calumbell (changes requested):
There is a lot in here. First up, colapsing the SASS into Tailwind classes looks like it has gone well. And it looks like the TS configuration is working properly now, lots of the annoying non-error TypeScript errors have vanished. Nice works.

There are still a few problems with the linter that I noticed:

1) The semi-colons are gone, could we put them back in? 

2) Single and double quotes appear to be handled differently in different context. IE. in the `ClassTable` component, single-quotes in the `<template>` tag were changed to double quotes, but not the in `<script>` tag. This happens is several (all?) other components and composables. My instinct is that this should be consistent across the codebase. Can you update the ESLint rule to enforce one or the other across the whole site?

calumbell:
@christophermiles Oh another thing! The branch you are pushing to here, `feature/v2-api`, has been merged to staging. Probably easiest if push this PR to `staging` also.

Thanks for your work!

christophermiles:
@calumbell I've fixed the semi-colons, apologies for the oversight.

With regard to the quotes, the Nuxt ESLint module uses the Vue recommended settings by default, which includes the use of double quotes in templates as the default: [https://eslint.vuejs.org/rules/html-quotes.html](https://eslint.vuejs.org/rules/html-quotes.html)

I have brought across any overrides we previously had in our ESLint config to the new config, and there doesn't seem to be anything explicitly in there about setting quotes to single in templates.

I'm not sure if I've ever seen a Vue project use single quotes in template blocks before, as there is often a need to do some string interpolation within attribute declaration, which (I assume?) would be made more difficult if using single quotes. 

But happy to override the default if there is some reason for using single quotes in the template code that I'm not aware of?

christophermiles:
@calumbell Unless with the quotes you meant stuff like this:

```
<<<<<<< HEAD
        <span>{{ col.value(data) ? "√" : "-" }}</span>
=======
        <span>{{ col.value(data) ? '√' : '-' }}</span>
>>>>>>> staging
```

In which case, yeah, I think it does make more sense for these to be single quote since these are JavaScript expressions. I thought you might have meant single quotes for the attributes.

I'll have a look to see if there's a setting for single quotes in JS expressions in templates.

christophermiles:
@calumbell Apologies for the delay, have made those changes now. 

I also made a change since the last PR to use the ['one true brace style' setting for the ESLint stylistic plugin](https://eslint.style/rules/js/brace-style#_1tbs), which was defaulting to ['stroustrup'](https://eslint.style/rules/js/brace-style#stroustrup)

It seemed like 'one true brace style' was closer to the convention being used (and is certainly closer to what I would normally use!)

christophermiles:
@calumbell @eepMoody I see one of the automated checks is failing though — I had a log at the logs and it seems to be something to do with some Python code?

calumbell (approved):
Hey @christophermiles, excuse the delay in getting this reviewed. I have been busy and this PR is a monster. Thanks for your contribution, but if we could keep future PRs a bit smaller and focused on single issues we'll be able to get these reviewed much faster.

The DX changes appear to be good! Poking around on local I encounter way fewer weird non-errors.

Many of the Tailwind changes are positive, but I have a few concerns (i'll slap these into their own issue, they are way beyond the scope of this PR). TLDR if we are using Tailwind, we should use be Tailwind classes and not the old classes copy-pasted from the SASS file to the Tailwind config. It buries them a bit to deeply and makes them (in my opinion) more difficult to maintain.

Thanks again for your contribution! Enough time has passed that we have some merge conflicts in our dependencies. Hopefully will only take a merge and a `npm i` to resolve

calumbell:
> @calumbell @eepMoody I see one of the automated checks is failing though — I had a log at the logs and it seems to be something to do with some Python code?

Yeah, the automated checks have been giving me problems as well. I also haven't gotten any useful feedback from them. I think we should get rid of them, what do you think @eepMoody ?

calumbell:
(Closing and reopening to re-run tests)

eepMoody:
> > @calumbell I think we should get rid of them, what do you think?

Yeah go for it! It was an experiment but it clearly isn't producing anything useful.

calumbell:
@christophermiles Apologies for how long this has taken to merge. I have submitted a PR to remove the AI code reviewer (#678), which is not working and blocking the merge. Once this is approved we should be able to finally merge these changes to `staging`.

eepMoody:
@calumbell do you want me to just override the protections and get this merged in? I hadn't ever intended that AI code review thing to be a blocker 🙏 sry

PR #615, opened 2025-01-09
Fix broken API link

joelshea:
Link was pointing at old URL that is no longer used. Updating to the correct repo.


calumbell (approved):
Good spot, thanks!

Issue #634, opened 2025-02-23
Refactor sidebar navigation into its own component


calumbell:
# Issue Description

While exploring options in how to resolve issue #507 by expanding the range of behaviours the website sidebar was capable of I found myself a overwhelmed by the `layouts/default.vue` file. In its current state, this layout component is not easily maintainable and would benefit from a refactor. For example: there is lots of UI programming logic implemented here that doesn't really have any buisness being in a layout file and which really bloats the component.

## Proposed Solution

There is a fair bit of work to do, but the piece of logic to refactor that would benefit implementing issue #507 the most would be to place the navigation menu into a new `<NavBar>` coponent. This would streamline the layout a great deal - both the JSX markup and the scripts.

The component shouldn't require any props, API classes to fetch the classes can be made within the component.

Issue #507, opened 2024-06-10
Feature: move subclasses to sub-tab


calumbell:
We would like to update the sidebar (implemented in `layouts/default.vue`) so that when a class is being viewed via the `/class/[id]` route, its subclasses are visible in the sidebar. Something like this:

![Image](https://github.com/open5e/open5e/assets/47407031/d0b414e6-e0d4-47b7-be02-d9a0e275230e)

It might be worth considering how we avoid allowing the sidebar to people excessively long. One idea might be to hide the other classes in the sidebar when one is already being viewed, but this might impair navigation too much.

calumbell:
Alright, so I have taken a look at getting this feature implemented, and here are my thoughts. Essay incoming.

**TLDR: probably best to wait until API V2 before undertaking this fix**. I'll tag this issue as blocked, but if you have a genius solution to solve it then be my guest!


## One Possible Solution

Adding the subclasses as a nested section to each class would require a refactor of how we currently handle the genration of navlinks. With possibly three-layers of nested links, recursive generation of subroutes seems like the wisest idea.

Assume we can format route data from the in the form:
```
type Route = {
  title: string,
  url: string,
  subroutes = Array<Route>;
}

```

Then the follow pseudo-code might be a good way to recursively generate the nav-links:

```
function generateLink(route: Route, baseUrl = '/') {
  if (!route.title || !route.url) return;
  return <li>
     <a :href="baseUrl + route.url">{{ route.title }}</a>
     <ul v-if="route.subroutes.length > 0">
        {{ subroutes.map((subroute) => generateLink(subroute, baseUrl + url)) }}
     </ul>
  </li>
}
```

This might be called in the component template like:

```
<template>
  ...
  // Navlinks
  <ul>
    {{ routes.map((route) => generateLink(route)) }}
  </ul>
</template>
```

(NB. this is pseudo-code and hasn't been tested)

## The Issue

The principal obstacle to this solution is fetching and formatting the API data in a way that avoids spaghetti while keeping the API response payloads as small as possible given how the Open5e API handles subclasses (archetypes in the parlance of the API)

Subclasses do not have their own endpoint, they are returned as an array for a given class. ie. Barbarian subclasses are access via the `archetypes` property returned by https://api.open5e.com/classes/barbarian/ . This property contains an array of archetype objects.

[...]

PR #635, opened 2025-02-23
Feature: `<NavBar>` component

calumbell:
This PR closes issue #634 by refactoring the navigation bar code from the `layouts/default.vue` file into the new `<NavBar>` component. This considerably streamlines the layout component, makingit more maintainable moving forward, and it also makes developing further features on the NavBar (ie. issue #507) simpler to develop.

Very little additional code was add this feature. It was mostly a straight cut and paste with a little tidying up. The change to the `composables/api.ts` was to fix a type error when passing a boolean prop to as a filter param to the `useFindMany()`.

### DIFF
--- a/composables/api.ts
+++ b/composables/api.ts
@@ -96,7 +96,7 @@ export const useAPI = () => {
 
 export const useFindMany = (
   endpoint: MaybeRef<string>,
-  params?: MaybeRef<Record<string, string | number>>
+  params?: MaybeRef<Record<string, string | number | boolean>>
 ) => {
   const { findMany } = useAPI();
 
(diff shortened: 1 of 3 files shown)


composables/api.ts
```diff
@@ -96,7 +96,7 @@ export const useAPI = () => {
 
 export const useFindMany = (
   endpoint: MaybeRef<string>,
-  params?: MaybeRef<Record<string, string | number>>
+  params?: MaybeRef<Record<string, string | number | boolean>>
```
calumbell, line 99: This change is techincally outside of the scope of this PR, but it was small and annoying enough that I fixed it here.

This fixes a type error that gets thrown when a boolean value is passed as a query param to the `useFindMany()` function. This handles the specific case of when we request our base classes by setting the `is_subclass` field to `false.

```
const { data: classes } = useFindMany(API_ENDPOINTS.classes, {
  fields: ['name', 'key'].join(),
  is_subclass: false,
});
```

eepMoody (approved):

Issue #633, opened 2025-02-22
Add creature Traits to `/monsters/[id]` page


calumbell:
# Task Description

Monster traits are missing from our monster pages, add them to the page (above actions)

**Observed Behaviour.** Monster traits returned by the API are not rendered on the frontend on the `/monsters/[id]` page.

**Desired Behaviour.** Monster traits should be visible on the `/monsters/[id]` page. They should be rendered immediately above the Actions section. If no traits are returned by the API for a given monster, the Traits heading should be omitted entirely.

If we look at the WotC SRD Goblin's page on Open5e `staging` branch (`/monsters/srd_goblin`) pointed at the Open5e API `staging` branch, we that the goblin statblock is missing its Traits:

<img width="700" alt="Image" src="https://github.com/user-attachments/assets/1a2097b7-fde1-4212-8e3e-eb997d752846" />

As of a recent developments on the API, monster traits are now returned by the o5e API:

`/v2/creatures/srd_goblin/?fields=name,traits` - Open5e-API 'staging' branch
```
{
    "name": "Goblin",
    "traits": [
        {
            "url": "http://localhost:8000/v2/creaturetraits/srd_goblin_nimble-escape/",
            "key": "srd_goblin_nimble-escape",
            "name": "Nimble Escape",
            "desc": "The goblin can take the Disengage or Hide action as a bonus action on each of its turns.",
            "type": null,
            "parent": "http://localhost:8000/v2/creatures/srd_goblin/"
        }
    ]
}
```

syhue:
can I take this?

calumbell:
@syhue Done!

PR #639, opened 2025-02-26
Feature: Add Creature Traits to Monster Detail Page

syhue:
This PR closes issue https://github.com/open5e/open5e/issues/633 by adding the creature traits section on top of the actions section in the monster detail page. 

Screenshots: 
image

image

### DIFF
--- a/pages/monsters/[id].vue
+++ b/pages/monsters/[id].vue
@@ -240,7 +240,7 @@
     </section>
 
     <!-- TRAITS -->
-    <section v-if="monster.traits.length !== 0">
+    <section v-if="monster.traits?.length !== 0">
       <h2>Traits</h2>
       <ul id="traits-list">
         <li v-for="trait in monster.traits" :key="trait.key" class="my-1">
--- a/pages/monsters/[id].vue
+++ b/pages/monsters/[id].vue
@@ -239,6 +239,19 @@
       </p>
     </section>
 
+    <!-- TRAITS -->
+    <section v-if="monster.traits.length !== 0">
+      <h2>Traits</h2>
+      <ul id="traits-list">
+        <li v-for="trait in monster.traits" :key="trait.key" class="my-1">
+          <span class="font-bold after:content-['._']">
+            {{ trait.name }}
+          </span>
+          <md-viewer :inline="true" :text="trait.desc" :use-roller="true" />
+        </li>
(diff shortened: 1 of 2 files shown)


calumbell (changes requested):
Welcome to Open5e!

Thanks for your contribution. The code you have added looks great, solves the issue as intended. No notes! However this PR isn't currently passing tests (`tests/unit/page/monster.test.tsx` specifically).

It looks like this is because our mock data doesn't currently include the `traits` field (V2 API is still a WIP, this field is new). So when we hit the following line (243 of `pages/monsters/[id].vue`):

```
<section v-if="monster.traits.length !== 0">
```

JavaScript throws an error because `traits` is undefined and undefined doesn't have a `length` property.

Simpliest solution I can think of would be to use optional chaining to check the length of the `traits` array. Something like `traits?.length`. 

We would also like to update the mock data on `tests/unit/page/monster.test.tsx` - adding the `traits` array to the JSON that the page component is tested against:

```
mockNuxtImport('useFindOne', () => {
  return () => ({
    data: {
       ...
       "traits": [
         {
              "url": "http://localhost:8000/v2/creaturetraits/srd_goblin_nimble-escape/",
              "key": "srd_goblin_nimble-escape",
              "name": "Nimble Escape",
              "desc": "The goblin can take the Disengage or Hide action as a bonus action on each of its turns.",
              "type": null,
              "parent": "http://localhost:8000/v2/creatures/srd_goblin/"
          }
       ],
       ...
    }
  })
});
```

But I think it would be reasonable to handle that in a seperate issue.

Thanks again for your contribution, let me know if you have any questions :)

syhue:
Hi, @calumbell 

Thank you so much for your detailed explanation. 

I have made a change about the optional chaining check and the adding of the testing data. 

https://github.com/open5e/open5e/pull/639/commits/d4427f520c761536fa272140e3742da2ab008b5b

Let me know if this is expected.

calumbell (approved):
Thanks for making those changes! Tests are passing and everything looks good on local. I'll need to close and re-open the PR to rerun validation and tests, but once that is done i'll go ahead and merge.

If you are keen to work on another issue, we have plenty and any further help would be greatly appreciated!

calumbell:
(Closing and reopening to rerun PR validation)

Issue #636, opened 2025-02-24
Hide Size column of `/monsters` on mobile


calumbell:
# Description

Currently, the results table on the `/monsters` route is looking pretty cluttered when viewed on at narrow screen widths. This will negatively impact the browsing experience for users on mobile.

<img width="380" alt="Image" src="https://github.com/user-attachments/assets/78902819-c4d6-4811-a7ee-b1b843e2cb60" />

*`/monsters` viewed at a 380px screen width*

The simplest solution to this issue would be to omit one of the columns when the page is viewed at narrow screen widths. I believe that dropping the `Size` column would be the best choice to leave off when screen space is limited (speaking from personal experience, monster CR and Type are by far the most important considerations for me when i am browsing for a monsters to use).

Elsewhere in our codebase we use the [TailwindCSS breakpoint selectors](https://tailwindcss.com/docs/responsive-design) to handle much of our responsive design. This is the technique that we should aim to use here.

## Task Overview

- Hide the `Size` column of the Monsters table at narrow screen widths
- Hide the `Size` filtering UI at the top of the Monsters tables at narrow screen widths

Bonus points if you can find a way to unset the Size filtering state on the Monsters table when the Size column is omitted

PR #641, opened 2025-02-27
Task: Hide Size Column of Monsters Overview on Mobile

syhue:
This PR closes issue https://github.com/open5e/open5e/issues/636. 

- Hide the Size column of the Monsters table at narrow screen widths. [Done]
- Hide the Size filtering UI at the top of the Monsters tables at narrow screen widths. [Done] 
- Bonus points if you can find a way to unset the Size filtering state on the Monsters table when the Size column is omitted. [x]

**Change Log:**
Adding a new flag `isLeastPriority` in the share components, if is true, it will be hidden on small screen. 

image

image

### DIFF
--- a/components/ApiResultRow.vue
+++ b/components/ApiResultRow.vue
@@ -1,7 +1,11 @@
 <template>
   <tr>
     <!-- Render each field defined in columns as a table cell -->
-    <td v-for="col in cols" :key="col.displayName">
+    <td
+      v-for="col in cols"
+      :key="col.displayName"
+      :class="{ 'hidden sm:block': col.isLeastPriority }"
+    >
       <template v-if="col.link">
         <span>
           <nuxt-link :to="col.link(data)">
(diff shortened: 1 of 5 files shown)


syhue:
@calumbell

Hi, I am new to Vue, I have been working on Angular. Could you share me some tips on the bonus part? 

I was thinking to use `ref`, `window.innerWidth` and function like `watch` to watch the inner width of the window constantly and clear the filter state at the right time. However, it do not work.

eepMoody (approved):

syhue:
@eepMoody 

Hm. That seems like a pretty reasonable solution. But it feels like you've identified a case where we should be solving this (checking for common breakpoints and setting some values to true/false) generally instead of doing one-off checks for specific things.

May I know the above is referring to which part? Do you mean hiding a column when the screen is small not really the best solution?

calumbell:
@syhue I believe he is refering to the bonus part. Your solution to hiding the column is solid.

A new PR has been created (#643) that will add the missing functionality to programatically query for screen size/breakpoint. It isn't ready to be merged to `staging` just yet (I have requested a small change to fix a bug), but when it is there should be no blockers for unsetting the table filters at small screen sizes

calumbell:
In the meantime, I think we are good to merge to `staging`. Thanks for your contribution!!

Issue #642, opened 2025-02-27
Create/find a utility that checks for md and sm screen widths, and can be used to set programmatic values


eepMoody:
Originally a bonus ask for #641 that didn't end up making sense to tackle locally.

The goal here would be to create or add some sort of function that checks whether the screen size is "at or below" a certain size, and can return programmatically useful values (eg. a boolean, or one of an enumerated list of "current sizes")

This could be in the store, or a method/function that is easy to import into other components. It would probably make sense to lean on a library or utility rather than rolling it ourselves.

Gabriel-Bueno32:
Hi I'd like to work on this

PR #643, opened 2025-02-27
feat: Add useBreakpoints composable for responsive design support

Gabriel-Bueno32:
This PR closes issue #642 

Adds a new composable function called `useBreakpoints` that provides support for responsive design. It allows the application to dynamically determine the current size of the viewport based on predefined breakpoints.

`useBreakpoints` examples return:

image

### DIFF
--- a/composables/useBreakpoints.ts
+++ b/composables/useBreakpoints.ts
@@ -9,13 +9,14 @@ const breakpoints = {
 };
 
 export function useBreakpoints() {
-  const width = ref(window.innerWidth);
+  const width = ref<number | null>(null);
 
   const updateWidth = () => {
     width.value = window.innerWidth;
   };
 
   onMounted(() => {
+    updateWidth();
     window.addEventListener('resize', updateWidth);
   });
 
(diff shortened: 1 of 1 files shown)


calumbell (changes requested):
Great work, fixes the issue and is laid out clearly. Thanks for your contribution!

While testing out your PR I did encounter a bug. It can be recreated by adding the following code to the page component for the index route:

```
<template>
     ....
     {{ currentSize }}
     ...
</template>

<script setup>
  ...
  const { currentSize } = useBreakpoints();
  ...
</script>
```

What we should see is this example is the current breakpoint for the screen size. However, when navigating to the page we hit a 500 error.

Screenshot 2025-02-28 at 08 33 21

Looking at the error, it looks like we are trying to access the `innerWidth` property of the `window` while the window is still undefined.

A fix for this solution would likely be to set the initial value of the `width` ref inside the `onMounted()` function. By the time this function runs, `window` will be defined. We will still want to declare the `width` ref outside of the method for scoping reasons.

Apart from the composable works perfectly (I triggered a HMR to test it)

calumbell (approved):
Thanks for getting those changes made! Bug is fixed, runs as intended on my computer, and all tests are passing.

I'll go ahead and approve and merge to `staging`. Thanks for your contribution!

calumbell:
(Closing and reopening to re-run PR validation)

PR #646, opened 2025-03-05
Task: Hide Size Column of Monsters Overview on Mobile v2 - fix the filter style

syhue:
This PR is the fix the side effect of PR #641.

BEFORE: 
image

AFTER:
image


calumbell (approved):
Good spot! Thanks for fixing this

PR #651, opened 2025-03-08
Fix newlines

### LINKED ISSUE #652: `<SourceTag>` causing errors on various pages
On the `staging` branch of the Open5e site (when pulling data from the `staging` branch of the Open5e API), errors are being thrown by the follow routes:

- `/feats/[id]`
- `/conditions/[id]`
- `/backgrounds/[id]`

(The same error is being thrown on all routes)

Image

This error message flags line 26 of `/feats/[id]`, which points at the code for computing the source key from the page URL:

```
// generate source key from page URL - for use with source-tab cmpnt
const sourceKey = computed(() => {
  if (!feat?.value?.document) return;
  return feat.value.document.url
    .split('/')
    .filter((exists) => exists)
    .pop();
});
```

However, following the merging of PR https://github.com/open5e/open5e-api/pull/643 on the API, the URL field is no longer returned as a subfield of the `document` field on these endpoints. This is throwing a value error.

The solution to this is to replace the `sourceKey` composable with the `document.key` field on the effected pages.

eepMoody:
Closes #652

Closes https://github.com/open5e/open5e-api/issues/646

### DIFF
--- a/components/MdViewer.vue
+++ b/components/MdViewer.vue
@@ -16,7 +16,6 @@
 
 <script setup>
 import { VueShowdown } from 'vue-showdown';
-import { watch } from 'vue';
 
 const props = defineProps({
   toc: { type: Boolean, default: true },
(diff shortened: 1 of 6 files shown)


eepMoody:
image

calumbell (approved):
Looks great! Didn't realise the fix was as simple as passing an additional prop to `<VueShowdown>`

Looks like you also fixed issue #652 while you were at it, nice!

Happy for this to be merged to `staging`, but it looks like you have it marked as a draft so not sure if I can just yet

PR #659, opened 2025-03-13
Sort Monster actions by type

### LINKED ISSUE #655: Sort monster actions by type on `monsters/[id]`
# The Issues

There are two problems on `/monsters/[id]` page in the page:

1. All the different types of actions (normal, bonus, reactions, legendary, lair) are currently listed under the **Actions** heading. We would like these to be organised by their action type
2. The code that handles the conditional rendering of the different action types is very verbose and difficult to maintain (Lines 243-343 of the page component for `/monsters/[id]`)

These two issues, while seperate, are being bundled here. Resolving point 1 without resolving point 2 will require a lot of additional code being written, which will be thrown out when point 2 is addressed.

## Reproduction

Spin up a test server and visit a statblock that you know to contain a mixutre of different action types. For example the Adult Bronze Dragon's (`/monsters/srd_adult-bronze-dragon`) statblock has its Legendary Actions (Detect, Wing Attack, and Tail Attack (the short one)) mixed in with its regular action list.

Image

The three legendary actions should be under their own `Legendary Actions` heading. As it is, it doesn't really make sense.

## One Possible Solution

Create a new `computed()` value that takes the `"actions"` field from the API response, and sorts each of the actions into an object where each key is an action type and each value is an array of the different actions of that type:

```
{
  "actions": [{...}, {...}],
  "bonusActions: [{...}],
  ...
}
```

(This could likely be achieved using the `reduce()` array method)

This `actions` can then be iterated over using the `v-for` directive to generate the various action type sections. Within each of these sections a second `v-for` directive can populate a list with the different actions that fall under that section.

In pseudo-code this might look like

```
<!-- THIS IS PSEDUO-CODE, DON'T EXPECT THIS TO WORK -->
<section 
  v-for="(actionType, actionsOfType) in actions" 
  :key="actionType"
/>
  <h2>{{ actionType }} </h2>
  <ul>
     <li v-for="action in actionsOfType" :key="action.name">
        <h3>{{ action.name }}</h3>
        <md-viewer inline="true" :text="action.desc" />
     </li>
  </ul>
</section>
```

calumbell:
Closes #655 

This PR splits out Monster actions by their type (bonus actions, reactions, etc.)

Screenshot 2025-03-13 at 09 01 38

Part of this fix included a refactor for how action types were conditionally rendered, which help to streamline the `/monsters/[id]` page component

### DIFF
--- a/pages/monsters/[id].vue
+++ b/pages/monsters/[id].vue
@@ -251,8 +251,7 @@
           <span class="font-bold after:content-['_']">{{ action.name }}</span>
           <span
             v-if="action.uses_type === 'RECHARGE_ON_ROLL'"
-            class="after:content-['_']"
-            class="cursor-pointer font-bold text-blood before:text-black before:content-['_('] after:text-black after:content-[')'] hover:text-black dark:before:text-white dark:after:text-white dark:hover:text-white"
+            class="cursor-pointer font-bold text-blood before:text-black before:content-['_('] after:text-black after:content-[')_'] hover:text-black dark:before:text-white dark:after:text-white dark:hover:text-white"
             @click="useDiceRoller('1d6+0')"
           >
             {{ `Recharge ${action.uses_param}-6` }}
--- a/pages/monsters/[id].vue
+++ b/pages/monsters/[id].vue
@@ -240,13 +240,12 @@
       </ul>
     </section>
 
+    <!-- CREATURE ACTIONS -->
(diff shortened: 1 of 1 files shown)


calumbell:
(Closing and reopening to re-run validation after fixing merge conflicts)

eepMoody (approved):

PR #684, opened 2025-04-17
Add fallback for initiative

eepMoody:
Added a fallback to dexterity, for monsters that don't have an initative_bonus already. This doesn't overwrite the incoming value if it exists, so it should be forward-compatible when we add calculated values to the API.

Currently most monsters on staging come back undefined for this:
image

With the fallback, they now fall back to dexterity bonus. 
image

### DIFF
--- a/composables/monsters.ts
+++ b/composables/monsters.ts
@@ -28,6 +28,11 @@ export const useMonster = (slug: string) => {
         modifier: useFormatModifier(monster[ability], { inputType: 'score' }),
         save: monster[`${ability}_save`],
       }));
+
+      // Add synthetic initiative bonus property
+      monster.initiative_bonus =
+        monster.initiative_bonus ?? monster.modifiers?.dexterity;
+
       return monster as Record<string, string>;
     },
   });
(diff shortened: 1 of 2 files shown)


calumbell (changes requested):
Hey, thanks for catching this. Can't believe I missed it!

One *tiny* thing that got lost in this PR is the `"+"` ordinal that precede positive or zero initiative bonuses. These can be easily added back in using the `useFormatModifier()` composable. Something like:

```
<dd @click="useDiceRoller(initiativeBonus)">
  {{ useFormatModifier(initiativeBonus) }}
</dd>
```

(Happy to add this in myself!)

Otherwise looks good, tests and build processes are passing locally. PR does what it says on the tin!

eepMoody:
@calumbell Oh yeah if you can see how to add it quick go for it! And feel free to merge it in if you think it's GTG

eepMoody:
Also it should kick off an auto-preview for testing purposes when you update it. No more needing to run local at all _Wheeeee!_

calumbell (approved):
The issue with the missing ordinal is fixed. Tests passing, all is looking good!

PR #685, opened 2025-04-17
Encounter builder

eepMoody:
## Description

This is a hefty PR but fairly focused in scope. It primarily creates three new related components and some supporting stores, to allow encounters to be built and managed in the app. 

On mobile this is all hidden for now, as it's fairly complex and I haven't figured out how to properly tackle it UX/UI wise.

[Preview Link](https://deploy-preview-685--open5e-preview.netlify.app/monsters)

### Adds an encounter tracker

This allows a monster to be added via any monster page, and does encounter math following 2014 rules. It allows a party to be defined for comparison then renders the difficulty using the gauge on the bottom

image

### Typeahead Monster Search
To add directly, the /search endpoint of the api was used to construct a rudimentary typeahead search. This fetches the monsters in batches of 25 until it builds the full list.

image

### Encounter Summary Button
When the encounter builder is closed, it persists in this button, which can be clicked to open it back up. This button shows the current encounter monster count and difficulty, so a user with a small screen can assemble monsters without having the builder open. It uses the same shared color coding as the main editor.

image
image

### Add/remove buttons
The mosnter list and detail pages have buttons to add and (if they are in the encounter) remove them. 

image

image

### Additional tweaks
There are a few more subtle additions, including prefetching for the monsters pages which accelerates load of secondary pages

## Related Issue
None, new development.

## How was this tested?

This was tested against the `staging` and `main` APIS. Manual QA tests include:
- Add a monster
- Remove a monster and add it again
- Increment/decrement a monster
- Search for a monster and add it
- Add a monster from the list page, then disable its source (should persist)
- Go to a monster page, disable that source, add the monster (works!)
- Add enough monsters to step through all difficulties with main builder
- Same but with summary view
- Search for "fire" in the combobox to confirm lazy loading fetches >25 entries
- Add a monster from the combobox
- Add an existing monster from the combobox to ensure it increments
- Adjust the number of party members to ensure encounter difficulty changes
- Same but with level
- Ensure a monster can be clicked-through to from the builder

Mobile
- Open on mobile and ensure no controls for the encounter builder render
- On monster details page
- On monster list page


components/ApiResultRow.vue
eepMoody, line 1: This allows a custom template to be injected into a row. I used this to add the new buttons to the monster table
image

components/MonsterAbilities.vue
```diff
@@ -72,7 +72,7 @@
 
 <script lang="ts" setup>
 import { useFormatModifier } from '@/composables/useFormatModifier';
-import { useDiceRoller } from '@/composables/useDicerRoller';
```
eepMoody, line 106: Just a typo here. It was "Dicer Roller" and I assumed that was unintentional, so updated it to "Dice Roller"

components/MonsterSearch.vue
eepMoody, line 1: This is a typeahead search box built on the Combobox headless UI component.

It's a bit gnarly because Headless doesn't cover all the basic features, but I based it on a handful of fairly common implementations found across the web and didn't hit too many snags.

components/PartyBuilder.vue
eepMoody, line 1: Subcomponent of the Encounter Builder. I assumed we might have other reasons to want to create/design a party (eg initiative tracker) so I abstracted this out

composables/useEncounter.ts
eepMoody, line 1: Store for the new encounter builder.

The basic principle is that when a monster is added from eg. a table it requires minimal display information plus an ID.

Then the store fetches and hydrates the full monster data (if it is a new monster), which is kept in localstorage. It is intentional that this stores multiple full monsters, as I intend to use that data to populate statblocks in the encounter manager later.

composables/useEncounter.ts
```diff
@@ -0,0 +1,206 @@
...
+  );
+  const monsterCache = useLocalStorage<Record<string, any>>(
+    'monster-cache',
+    {}
+  );
+  const { get } = useAPI();
+  const { partyXPBudget } = usePartyStore();
+  const xpCalculator = useXPCalculator();
+
+  const difficultyColors: Record<DifficultyLevel, string> = {
```
eepMoody, line 40: I wasn't sure if this was the best place to put this, but it felt reasonable since nothing but encounters ever uses encounter difficulty.

This isn't used for 2024 rule tho, so I may refactor them out later.

composables/useFindPaginated.ts
eepMoody, line 1: These changes are somewhat unrelated, but add prefetch to the monster table so the next page loads a bit faster.

composables/useParty.ts
eepMoody, line 1: Abstracted out the party storage in case we want to use it in a different context from encounters

composables/useXPCalculator.ts
eepMoody, line 1: Separated XP rules since it looks like 2024 uses different rules. Wanted to be able to keep those separate.

layouts/default.vue
```diff
@@ -6,7 +6,7 @@
   >
     <sources-modal :show="showModal" @close="showModal = false" />
     <div
-      class="bg-dark m-auto grid h-full min-h-screen max-w-[1280px] grid-flow-col transition-all sm:ml-0 sm:w-screen sm:grid-cols-[14rem_1fr] sm:overflow-y-auto sm:transition-none"
+      class="bg-dark m-auto grid h-full min-h-screen max-w-[1440px] grid-flow-col transition-all sm:ml-0 sm:w-screen sm:grid-cols-[14rem_1fr] sm:overflow-y-auto sm:transition-none"
```
eepMoody, line 12: Made it a little wider to account for the sidebar on wider screens

pages/monsters/[id].vue
```diff
@@ -4,19 +4,30 @@
     <div class="flex items-end justify-between gap-8">
       <h1 class="flex-auto">{{ monster.name }}</h1>
 
-      <div class="flex flex-none items-start">
-        <button
-          class="flex items-center gap-1 rounded-md p-1 text-xs text-blood outline outline-1 outline-blood hover:bg-blood hover:text-white"
-          @click="toggleMode()"
-        >
+      <div class="flex flex-none items-start gap-2">
+        <!-- <button           class="flex items-center gap-1 rounded-md p-1 text-xs text-blood outline outline-1 outline-blood hover:bg-blood hover:text-white"
```
Comment (eepMoody) [inline pages/monsters/[id].vue:13]: Currently Compact Mode does not work in stating/prod, so I've temporarily hidden the button here

pages/monsters/[id].vue
```diff
@@ -4,19 +4,30 @@
...
           <icon
             :name="
               mode === 'compact'
                 ? 'heroicons:arrows-pointing-out'
                 : 'heroicons:arrows-pointing-in'
             "
           />
-          {{ mode === 'compact' ? 'Regular statblock' : 'Compact statblock' }}
+          {{ mode === 'compact' ? 'Regular statblock' : 'Compact statblock' }}</button> -->
+        <button
```
Comment (eepMoody) [inline pages/monsters/[id].vue:23]: New buttons to pop monsters into the encounter

pages/monsters/[id].vue
```diff
@@ -40,7 +51,11 @@
         {{ monster.alignment }}
       </span>
 
-      <source-tag :title="monster.document.name" :text="monster.document.key" />
+      <source-tag
```
Comment (eepMoody) [inline pages/monsters/[id].vue]: I have some plans here, just adding some additional info to enable that

pages/monsters/index.vue
```diff
@@ -87,6 +87,43 @@
           sortValue: 'size',
           isLeastPriority: true,
         },
+        {
```
eepMoody, line 97: New add/remove buttons as custom component

pages/monsters/index.vue
```diff
@@ -143,4 +187,28 @@ const {
   prevPage,
   nextPage,
 } = paginator;
+
+const encounterStore = useEncounterStore();
+
+const addToEncounter = (monster: any) => {
```
eepMoody: Don't love that this needs to be defined on each page but wasn't sure how best to abstract it out.

tailwind.config.ts
```diff
@@ -48,6 +48,21 @@ module.exports = {
 
       //adds z-index from 1 to 100
       zIndex: zIndex,
+
```
eepMoody, line 54: RIDE IN TO THE _DANGER ZONE!!!!_

tests/unit/components/EncounterBuilder.test.tsx
eepMoody, line 1: I did some automated test-building with AI on this one, then hand-modified. Because writing tests sucks when it comes to the store.Tests that the component renders, that monsters can be added, that the state properly reflects whether a monster is there, and that the various management calls are still firing.

types/monster.ts
eepMoody, line 1: I'm not sure if I should call this one something specific, but it seemed like a reasonable starting point for an internal UI-layer concept of "A Monster". Just the basic you'd need to render it in-place, plus the ID so you can fetch the rest.

calumbell (changes requested):
Hey, first off wanted to say that this looks great and seems to be working really well!

A did spot a little bug with the XP calculator spitting out a `NaN` result when PCs of 20th level are included in the encounter calculation:

ScreenRecording2025-04-20at09 41 31-ezgif com-video-to-gif-converter

Not necessarily a bug, but the custom template that gets passed to the `<ApiResultsTable>` component on the `/monsters` page feels a little unwieldly. I am unfamiliar with the API you have used here, but it looks like an inline component declaration? Any idea if it be possible to move this into its own component file to improve readibility of the `/monsters` page markup? This would be a cherry on top

eepMoody:
Nah that's just an Easter egg, nobody actually reaches level 20, right? 😛

Seriously though good catch! I'll get that fixed.

I'll give it a try on the table. It was giving me some trouble and doing an inline component/render function is how I've usually solved that issue in React land. But maybe I just missed something.

calumbell (approved):

PR #708, opened 2025-05-25
[REFACTOR] Removed references `depth` query parameter 

### LINKED ISSUE #706: [REFACTOR] Remove references to `depth` parameter in API calls
## Description

With our new and improved data serializers on the V2 API, the `depth` query parameter is no longer needed to fetch nested ForeignKeys. Passing a `depth` query parameter in fact has *no* effect on the V2 API, and references to it should be removed from the front-end to avoid creating the false impression that it actually does do something.

For example, the default page Layout makes a request to the  `/v2/documents` endpoint to fetch a list of sources to populate the Source Selector with:

```
const { data: documents } = useDocuments({
  fields: 'none',
  depth: 0,
});
```

The `depth` parameter does nothing here. This can be safely refactored as:

```
const { data: documents } = useDocuments({ fields: 'none' });
```

The process should be repeated whenever there is a `depth` parameter included in an API call (there is a good chance that some `depth` parameters are being inserted in queries within the Composable itself)

## Desired Behaviour
Remove the `depth` query parameter from *all* queries made by the FE to the API

calumbell:
## Description

This PR performs a little bit of housekeeping. 

The `depth` query parameter was used by the V1 API to help handle the Serialization of nested FK fields. For the V2 API this parameter is no longer needed. Passing a `depth` query parameter has zero effect on the API result, so it has been removed to avoid any confusion moving forward.

## Related Issue

Closes #706 

## How was this tested?

- Manually tested using the Nuxt development server (`npm run dev`)
- `npm run build`
- `npm run test`


eepMoody (approved):
Good catch, and definitely glad we've updated to not require this in so many places

calumbell:
Merge conflicts fixed. Closing and re-opening to re-run validation

PR #711, opened 2025-05-28
[REFACTOR / DX] Rename tightly-coupled components to improve alphabetical sort order

### LINKED ISSUE #710: [REFACTOR] Rename tightly-coupled components to improve alphabetical sort order
## Problem Description

Navigating between related, tightly-coupled components in the `/components` directory is more difficult than it should because conceptually related components are not always adjacent to each other due to their alphabetical sorting order.

## Proposed Solution

My first instinct was do group component-groups into subdirectory, but  [Vue Style Guide](https://vuejs.org/style-guide/rules-strongly-recommended.html#tightly-coupled-component-names) advices against this, instead recommending:

*Child components that are tightly coupled with their parent should include the parent component name as a prefix. If a component only makes sense in the context of a single parent component, that relationship should be evident in its name. Since editors typically organize files alphabetically, this also keeps these related files next to each other*

It would be a good idea to implement a similar feature in our component library, especially for the components that are used to generate the API Results Table or the Encounter Builder, which both contain numerous sub-components.

calumbell:
## Description

This PR renames a number of component so that closely related and tightly-coupled components appear adjacent to each other when sorted alphabetically (as suggested in the  [Vue Style Guide](https://vuejs.org/style-guide/rules-strongly-recommended.html#tightly-coupled-component-names)).

Component 'families' that were grouped together include:

- Modal Menus (Modal, ModalReportIssue, ModalSourceSelector)
- API Results Table (ResultsTable, ResultsTableFilter, ResultsTableHeader, ResultsTablePaginator, ResultsTablePaginatorButton, ResultsTableRow)
- Encounter Builder (EncounterBuilder, EncounterBuilderMonsterSearch, EncounterBuilderPartyBuilder, EncounterBuilderSummary)
- Navigation (NavMenu, NavMenuLink)

This makes the components directory much easier to traverse, and should improve maintainability

## Related Issue

Closes #710


eepMoody:
Can we do a merge from staging once all the other changes are in, before we do this one? There's enough wonky stuff that I'm having trouble testing it, but I _think_ it's all due to issues corrected elsewhere.

tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "./.nuxt/tsconfig.json"
+}
```
calumbell: Apologies if I am being dense, but I am not 100% what this is doing in this PR. Would you be able to take a moment to explain the intention behind adding the tsconfig here?

Asking because this directly conflicts with the tsconfig introduced in PR #712, and it would be helpful to know exactly what it is solving so I can merge it in correctly. Thanks in advance!

eepMoody: @calumbell no that's my bad, something weird was going on with my local docker build. I was just gonna try and do a merge and didn't even mean to push this up. Let me take it out. 🙇 sorry!

eepMoody: I have a feeling it's because I switched from the branch where you'd set up the tsconfig directly to this and something got cached funky.

eepMoody (approved):
Tested every page and as many functionalities as I could, looks solid to me!

eepMoody:
Though something funky does seem to be happening with the build check here too, been sitting for a while. 
🤔  not sure what's up. I'm willing to chalk it up to something being flaky and merge it in anyways to see what happens?

eepMoody:
Yeah seems to be fine, same as what happened in my local once I tore it all down and ran it again 🤷

PR #712, opened 2025-05-29
[BUGFIX / DX] Configured Typescript

### LINKED ISSUE #709: [BUG] Typescript not configured correctly
## Description

VSCode is currently flagging a *lot* of Typescript errors. These errors don't appear to be interacting with our linter in any meaningful way, so it is difficult to get visibility on them. It would be good if we could update date our Typescript configuration so that these errors are A) visible to the ESLint, and B) resolved.

## Observed Behaviour

VSCode raises many type errors in the Open5e repo. These errors don't appear to interact with the Linter. Running `npx tsc --noEmit` doesn't perform any type checking.

## Desired Behaviour

Fix all type errors being thrown by VSCode. Update TS configuration so that type-checking can be performed at the command line. If possible, add typechecking to the linting process.

## Context

My initial investigation into the issue turned up the following:

No type checking occured when `npx tsc --noEmit` was run in the terminal. Instead we see a 'how to use . Poke around online led me to believe that this was because there is no`tsconfig.json` file.

 I ran `npx tsc --init` to generate a `tsconfig.json` file. `npx tsc --noEmit` now executes and flags 144 TS errors across 29 files (yikes)

calumbell:
## Description

This PR addresses an issue where Typescript wasn't correctly configured on the `staging` branch. This has been resolved by adding a `tsconfig.json` file, which is required by `tsc` to check types.

The remainder and vast majority of this PR was devoted to fixing the hundred or so type errors thrown up by the TS compiler. Most of these were very simple fixes; missing import statements, mis-typed function parameters, etc. Some were much more involved - moving deprecated Nuxt2 options out of the Nuxt config to different configuration files and plugins.

N.B. that the TS Compiler is not able to parse `.vue`. The [vue-tsc package](https://www.npmjs.com/package/vue-tsc) is required to do this. My feeling that this should be handled in its own issue, as this PR is already quite bloated.

## Related Issue

Closes #709 

## How was this tested?

- Spot testings on the test server (`npm run dev`) pointed at the `staging` branch of the API, also running on a test server
- `npm run build` (build process completes without issue)


eepMoody:
Nice. I made an issue #714 to keep track of the vue-tsc thing, so we don't forget about it!

tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    "types": ["@types/node", "@nuxt/types"],
+    "esModuleInterop": true,
+    "noEmit": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
+    "skipLibCheck": true
+  },
+  "include": [
+    "nuxt.config.ts",
+    "pages/**/*.vue",
```
eepMoody: Just as a thought, should we leave these out since they're not actually being parsed yet? (Or make some sort of note/comment in the file for anyone who's looking at it down the road?)

eepMoody (approved):

tsconfig.json
calumbell: Good idea. I'll need to make another commit to fix some merge conflicts so I can sort this then. Will likely comment it out and add a short explaination why it is commented out

calumbell:
Fixed merge conflicts. Closing and re-opening PR to re-run validation

PR #738, opened 2025-07-23
Source Selector Refactor

### LINKED ISSUE #736: [REFACTOR] Move Source Selection Button into `ModalSourceSelector` component
## Description

The Layout component (`/layouts/default.vue`) currently contains implementation code for the UI that opens the Source Selection modal and populates the tooltip with sources information. This is beyond of the scope of what a layout component should be expected to handle, and makes actually updating the layout difficult.

## Ideal Solution
Move the logic used to render the open source selector button from the layout component into the `ModalSourceSelection` component.

The `ModalReportIssue` component can be used as a model. In that the button for opening the menu and the modal that are opened are both children to the same parent-container.

calumbell:
## Description

This PR moves all of the logic/implementation details for the source selection modal out of the Layout and into a new `SourceSelector` component (which also encapsulates the `ModalSourceSelector` component, renamed here to `SourceSelectorModal`).

Screenshot 2025-07-23 at 23 06 03

The `SourceSelector` button has had its styles tweaked a little to make it match the Search Bar a little closer. A big restyle of the source selector button is probably overdue as it isn't immediately obvious looking at it that it is a piece of UI that you click on.

## Related Issue

Closes #736 

## How was this tested?
- Spot checked on development server (`npm run dev`)
- All tests passing (`npm run test`)
- Build process completes without error (`npm run build`)
- PR validation is passing


eepMoody:
That's a good point about the visual design of the selector. The whole site could probably use a bit of a visual refresh, maybe that's a good project to tackle for a v2 launch...

I'll noodle on it in Figma a bit, see if maybe there's some cool ways to incorporate some of the other features we've added over the years!

eepMoody (approved):

PR #740, opened 2025-07-24
Search Bar Refactor

### LINKED ISSUE #733: [REFACTOR] Move searching logic from Layout into `SearchBar` component
## Describing the Problem
Currently, the default page layout (`/layouts/default.vue`) contains all the implementation code for the search bar markup and fetching util. Really, the layout's job is to handle layout, and having a file that is so multi-use makes it difficult to update or maintain.

## Ideal Solution
It would make the Layout much easier to work with if all the implementation code for the search bar were moved into a new `SearchBar` component. This component should capture all of the markup included in the layout's `template` (lines 53-70) that renderers the bar, and replace it with a `<SearchBar />` component.

The declaration of the `doSearch()` method should also be captured by the component

calumbell:
## Description

This PR addresses a issue where implementation code for the search bar embedded in the site sidebar was placed directly in the layout folder. The problem was resolved by moving the code in question to its own `SearchBar.vue` component.

Visually there should be no changes, excluding the addition of an ellipsis at the end of the search bar text.

## Related Issue

Closes #733 

## How was this tested?
- Spot checked on Nuxt development server (`npm run dev`)
- All tests green (`npm run test`)
- Build process completes without error (`npm run build`)

### DIFF
--- a/components/SearchBar.vue
+++ b/components/SearchBar.vue
@@ -1,5 +1,4 @@
 <template>
-  <!-- SEARCH BAR -->
   <div class="relative">
     <div
       class="absolute inset-y-0 right-0 flex cursor-pointer items-center pr-2"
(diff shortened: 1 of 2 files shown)


calumbell:
This PR has been superceded by #744, which includes all the changes outlined in this PR

PR #742, opened 2025-07-27
fix: app fails when magic item rarity is null

### LINKED ISSUE #741: [BUG] `/magic-item/[id]` page crashes when `rarity` is null
## Description

When visiting the `/magic-item/[id]` page for an item that has a `null` value for its `rarity` property, the following error is thrown.

`Uncaught (in promise) TypeError: can't access property "name", $setup.item.rarity is null`

Which implies that Nuxt is trying to index into the `rarity` property to get its name, only to hit a key error because `rarity` is null. The magic item that alerted me to this bug was the 2024 SRD Belt of Giant Strength (`srd-2024_belt-of-giant-strength`). Examining the API payload in the network tab when visiting `/monsters/srd-2024_belt-of-giant-strength` I was able to confirm that the `rarity` field was null.

The fix for this is as simple as a null-check, optional chaining, or nullish coalseasing.

## To Reproduce

This bug was observed on the `staging` branch of the website running on a local Nuxt development server, pulling data from a fork of the `staging` branch of the API, running on a local Django developement server.

1. Visit a the single-view page for a magic item where the rarity is null (ie. 2024 SRD Belt of Giant Strength).

2. Note that the page doesn’t load. Viewing the JS console you see the error message described above.

ernestorbemx:
## Description

I added a null checking for the rarity properties in the `/magic-items/[id]` route

## Related Issue

Closes #741 

## How was this tested?
I edited the `magic-items/srd-2024_adamantine-armor` entry in my SQLite DB, setting 
rarity_id = NULL and loaded the page

image

### DIFF
--- a/app/pages/magic-items/[id].vue
+++ b/app/pages/magic-items/[id].vue
@@ -6,7 +6,7 @@
       </h1>
       <p>
         <em>
-          {{ item.category.name }}, {{ item.rarity.name }}
+          {{ item.category.name }}, {{ item.rarity?.name ?? 'Rarity unspecified' }}
           <span v-show="item.requires_attunement">
             ({{ 'requires attunement' }})
           </span>


calumbell (approved):
Hi, thanks for your contribution. Works perfectly, all tests are passing. I'll approve and get it merged to `staging`

PR #744, opened 2025-07-30
Website UI Refactor

calumbell:
## Description

This PR refreshes and refactors much of the site-wide UI that was looking a little tired.

This is a fairly substantial PR, but the main changes can be summarised as follows:

- Added a slender 3rd column to the page layout. The column acts as a vertical UI ribbon, gathering together sundry pieces of non-navigation UI into one place (source selector, dark mode toggle, etc.).
- Restyled the Navigation menu.
- Moving lots of UI implementation code out of the default layout file and into their own components. This way the layout can focus solely on the page layout.

Many pages and components required tweaking so that they played well with the changes made to `app/layouts/default.vue`. This mostly involved the tweaking of TaillwindCSS classes.

## Preview

[Deploy preview available here](https://deploy-preview-744--open5e-preview.netlify.app/)


app/components/ModalReportIssue.vue
```diff
@@ -42,151 +52,126 @@ const submitIssue = async () => {
...
-      :show="isOpen"
-      @close="isOpen = false"
-    >
-      <slot>
-        <!-- Bug submission form -->
-        <form
-          v-if="['ready', 'pending'].includes(status)"
-          id="report-issue-form"
-          method="POST"
-          :action="WEBAPP_URL"
```
calumbell, line 77: The `ModalReportIssue` component used to contain both the Report Issue pop-up for and the button to open it. This has been refactored so that the `ModalReportIssue` component only hanldes the pop-up menu and the new `ToolButtonReportIssue` component acts as the toggle that opens and closes the menu.

The advantage of this approach is that; A) it hews-closer to single-use component principals, making it more maintainable, B) it is more flexible as the modal menu can be opened without the `ToolButtonReportIssue`, if necessary

app/components/NavMenuLink.vue
calumbell, line 1: Component no longer needed, this logic is now handled in the `Navigation`component

app/components/Navigation.vue
```diff
@@ -0,0 +1,120 @@
+<template>
```
calumbell, line 1: This is a new component replaces the `NavMenu` component, but also captures the logo and search bar in the left navigation menu

app/components/SearchBar.vue
```diff
@@ -0,0 +1,66 @@
+<script>
```
calumbell, line 1: The `SearchBar` component contains code refactored out from the Layout. It contains logic and implementation code that doesn't really make sense to keep in the Layout.

app/components/ToolButton.vue
```diff
@@ -0,0 +1,25 @@
+<template>
```
calumbell, line 1: Think of this as an `abstract` component. It is used as a wrapper by the other Toolbar buttons using Vue's slots system to keep our code DRY and our button styles uniform.

app/layouts/default.vue
```diff
@@ -2,115 +2,71 @@
   <!-- BACKGROUND (visible behind page content at wide screen widths)       -->
```
calumbell, line 2: The default layout has been savaged in this PR. My goal with this refactor was to take much of the nuts-and-bolts implementation code out of the Layout and move it our components. THis is because it is my feeling that the layout should handle, well, the layout, and with all the implementation code in this file this was becoming difficult!

app/pages/backgrounds/index.vue
```diff
@@ -1,12 +1,9 @@
 <template>
```
calumbell, line 1: These changes were initially undertaken to fix x-overflow issues on this page that occured when the layout was updated, but they ended up fixing some faulty filtering functionality. Ibid `/classes`, `/conditions`, `/equipment`, `/feats`, `/rules` and `/species`

app/components/NavMenu.vue
calumbell, line 1: This component was renamed to `Navigation.vue`, but I suppose the refactor was substantial enough that GH doesn't recognise it

app/components/AppLogo.vue
calumbell, line 1: The component has been unused for many years

app/components/ToolBar.vue
calumbell, line 1: This component acts as the UI container for the new ToolButtons. The list could we be populated via a `v-for` directive if we wanted its content to be data-driven, but that felt like an over-engineered solution at this point.

calumbell:
CLosing and re-opening to re-run PR validation

eepMoody (changes requested):
Found one small issue in this one, but otherwise it seems to be good! I've tested the various list and item pages, submitted a test issue report, searched, etc.

The only bug I found was when attempting to open the list of sources after selecting some and returning to the site, I think there's some sort of issue with handling the list of selected items on open.

```
js:14 TypeError: t.filter is not a function
...
js:14 TypeError: Cannot read properties of undefined (reading 'length')
...
```

eepMoody:
One other, the "compact statblock" throws an error as well. But it's not a widely used feature, so I think it's equally valid to remove the link or to fix the issue.

```
DCOs2K7D.js:14 TypeError: Cannot read properties of undefined (reading 'split')
...
```

calumbell:
@eepMoody Thanks for catching that! This issue wasn't showing up on local, but happens clear as day on the test deploy. It looks like the Sources modal is trying to filter the sources while they are still undefined, so hopefully adding some type-guards might help.

I think we are good to remove the compact stat-block feature. I am not 100% sure what purpose it serves, and it means we need to maintain two separate layouts of the Monster page.

calumbell:
Thanks for your feedback @eepMoody. Looking at the preview build it looks like the issues you found are resolved:

- Source Selector now open correctly on page reload (added a few guards clauses and a missing deferencing step in the filtering by gamesystem logic)
- Removed the compact monster statblock links

calumbell:
Closing and re-opening to re-run PR validation

eepMoody (approved):
LGTM with those tweaks!

PR #746, opened 2025-08-09
Add Monster illustrations to /monsters/[id] page

### LINKED ISSUE #739: Add Monster illustrations to `/monsters/[id]` page
## Description
WIth PR https://github.com/open5e/open5e-api/pull/767 merged to `staging`, we now have access to monster illustrations via the `illustration` property.

There aren't many of the (yet!), but it would be good to display the few that we have on the `/monster/[id]` page for the relevant monsters

## Desired Solution

Add markup to the `/monster/[id]` that will render a monster image if the `illustration` field isn't null

na0kk0:
## Description

I modified the monster/[id] page to use the field illustration if it exists (also an alt attribute with the monster's name). 

## Related Issue

*Closes #739*

## How was this tested?

I checked with two monster records, one with a null illustration field and one with a valid value. 
image

image

### DIFF
--- a/app/pages/monsters/[id].vue
+++ b/app/pages/monsters/[id].vue
@@ -38,8 +38,9 @@
     </div>
 
     <img
-      v-if="mode !== 'compact' && monster.img_main"
-      :src="monster.img_main"
+      v-if="mode !== 'compact' && monster.illustration"
+      :src="useRuntimeConfig().public.apiUrl + monster.illustration.file_url"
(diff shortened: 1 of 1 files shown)


calumbell (approved):
Looks like everything is working as expected. Spot check on the test server looks good, all tests passing and build process completes without error.

I'll go ahead and merge this to `staging`, thanks for your contribution!

PR #759, opened 2025-09-05
[BUGFIX / TYPESCRIPT] Refactor `useFormatModifier` and `useDiceRoller` composables

### LINKED ISSUE #758: [BUG] Refactor `useDiceRoller` and `useModifier` to return obj. to destructure
## Describe the Bug
I encountered this bug while trying to fix sundry Typescript errors on the `/monsters/[id]` page:

```
<dt class="font-bold after:content-['_']">
  Initiative Bonus
</dt>
<dd
  class="w-min cursor-pointer font-bold text-blood hover:text-black dark:hover:text-fog"
  @click="useDiceRoller(initiativeBonus, {
    title: 'Initiative',
    subtitle: monster.name
  })"
>
  {{ useFormatModifier(initiativeBonus) }}
</dd>
```

The use of `useDicerRoller` and `useFormatModifier` is throwing a type-errors because `vue-tsc` is not recognising the auto-imports. When digging into this I discovered that this weird behaviour was happening because these composables were not being used in the correct way. Composables shouldn't be called directly in the `<template>` tag, and belong in the `<script setup>` tag.

The most ideomatic solution to this issue is to have the composables return an object that can be destructured containing these functions as a value. ie.

```
const { rollDice } = useDiceRoller();
const { formatModifier } = useFormatModifier();
``` 

These destructured functions can be used in the `<template>` tag not problem

## Desired Solution

`useDiceRoller` and `useFormatModifier` composables should be refactored so that the `rollDice` and `formatModifer` functions can be destructured from them in the `<script setup>` tag. All instances of the `useDiceRoller` and `useFormatModifier` will then need to be updated to work with the new component structure.

calumbell:
## Description

This PR refactors the `useFormatModifier()` and `useDiceRoller()` composables so that they return an object containing a function which performs that composables function. The purpose of this change was so avoid the composable being called directly in the `<template>` tag, which is goes against Nuxt best practices and also throws up a heap of Typescript errors

## Related Issue

Closes #758

## How was this tested?
- Spot checking on the Nuxt development server (pulling data from the `staging` branch of the API)
- `npm run test` (all tests passing)

## Build Preview
https://deploy-preview-759--open5e-preview.netlify.app/

### DIFF
--- a/app/components/InlineRoller.global.vue
+++ b/app/components/InlineRoller.global.vue
@@ -28,5 +28,5 @@ defineProps({
   signature: { type: String, default: '' },
 });
 
-const { rollDice } = useDiceRoller();
+const rollDice = useDiceRoller();
 </script>
--- a/app/components/InlineRoller.global.vue
+++ b/app/components/InlineRoller.global.vue
@@ -17,7 +17,7 @@
 <template>
   <span
     class="cursor-pointer font-bold text-blood hover:text-black dark:hover:text-fog"
-    @click="useDiceRoller(signature)"
+    @click="rollDice(signature)"
   >
     <slot />
   </span>
@@ -27,4 +27,6 @@
 defineProps({
   signature: { type: String, default: '' },
 });
+
+const { rollDice } = useDiceRoller();
 </script>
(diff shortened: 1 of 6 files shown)


calumbell:
Closing and re-opening PR to re-run validation

eepMoody:
Very slick refactor! Love the subtle updates, but much cleaner now you've taken this pass at it.

eepMoody (approved):

PR #764, opened 2025-09-17
[BUGFIX] Fix y-overflow layout shift

### LINKED ISSUE #737: [BUG] Y-scroll bar pushes off-screen at smaller screen widths
## Description
At 'smaller' screen width (< 1440px) where the main content spans the entire screen, there is a strange interaction between the vertical window scroll and the navigation menu on the left of the screen.

Image

When the vertical scroll bar is visible, the navigation menu is shifted off screen by the width of the bar. This is quite jarring when navigating between a page that overflows the page height and one that doesn't (as in the above video).

## Reproduction
This bug can be observed on the `staging` branch.
1. Spin up a development server (`npm run dev`)
2. Visit the home page of the development server with a browser window with a width less than 1440px (I tested at 1000px). Set the screen height so that the entire navigation bar is visible on screen.
3. Navigate to a page that overflows the screen in the Y-axis (ie. `/spells`). Make sure you can see the scroll bar.
4. Navigate to the top spell in the spell list (Abhorrent Apparition). Notice that this page doesn't overflow vertically. The bar disappears and the navigation menu shifts. 

## Expected Behaviour
The vertical scroll bar should have no effect on the position of the navigation menu.

## Hardware
This error was observed on the Firefox browser.

calumbell:
## Description

This PR fixes a visual bug where navigating from a page that didn't vertically overflow the browser window to one that did (and vice versa) would cause a noticable shift in the layout, most prominently in the position of the left navigation menu (which would be push off the screen when the y-scrollbar was visible).

The solution to this was to add `scrollbar-gutter` CSS classes to the O5e CSS reset.  This worked at large screen widths but absolutely wrecked havok on the UX at small screen widths, the sliding menus were not working as intended. This led to a refactor of the layout component so that collapsing menus were repositioned using `translate` classes instead of `margin` classes.

## Related Issue

Closes #737 

## How was this tested?

- Spot checked on the Nuxt development server (pulling data from the `staging` branch of the API running on a local Django development server)
- `npm run build` was run to confirm that the build process completed without error
- `npm run test` was run to ensure that all tests were still passing


eepMoody (approved):

calumbell:
Fixed merge conflict. Closing and re-opening PR to re-run validation

PR #765, opened 2025-09-17
fix: source selector modals getting out of sync

### LINKED ISSUE #762: [BUG] Source Selector Modal menus are not perfectly in sync
## Description

The Source Selection modal interface is not perfectly in-sync with itself. The menu can be opened in two different places (in the SearchBar on the left of the screen, and in the Tool ribbon on the right). If the Sources are updated in one, then the changes will not propagate into the other without a page reload.

Image

## To Reproduce
1. Spin up a local dev server running the `staging` branch of Open5e (pulling data from the `staging` branch of the API)
2. Navigate to the home page, note the two different pieces of UI that can be pressed to access the Source Selector Modal
3. Click on one of the Source Selector buttons to open the Source Selector modal menu. Change it (to anything) and click the "Update" button. This will close the modal menu.
4. Notice that the selected sources have been successfully updated.
5. Now click on the OTHER Source Selector piece of UI on the other side of the screen. This will open the Modal Menu again.
6. Notice that the checkboxes ticekd will not match those in the menu opened by the other piece of Source Selection UI. It will display the source selection as it was on page-load.

## Desired Behaviour
Whenever the selected sources state is updated, all instances of the Source Selection modal menu should update their state to match the app-wide state.

JeremyMees:
## Description
This PR resolves an issue where the source select modals were getting out of sync.

## Related Issue
Closes #762

## How was this tested?
Since there were no existing automated tests covering this bug, I verified the fix manually:
• Reproduced the issue by following the same steps shown in the video attached to the GitHub issue.
• Confirmed the bug is resolved after applying the fix.
• Ran npm run build to ensure the build process completed successfully without errors.
• Ran npm run test to confirm that all existing tests still pass.

### DIFF
--- a/app/components/ModalSourceSelector.vue
+++ b/app/components/ModalSourceSelector.vue
@@ -177,9 +177,12 @@ const { data: documents } = useDocuments({
 });
 
 const { sources, setSources, gameSystem, setGameSystem } = useSourcesList();
-const selectedSources = ref(sources.value);
+const selectedSources = ref([]);
 const emit = defineEmits(['close']);
 const closeModal = () => emit('close');
+
+// Keep selectedSources in sync with global sources state
+watchEffect(() => selectedSources.value = [...sources.value]);
 
 // filter documents by the current game system
 const documentsInSystem = computed(() => {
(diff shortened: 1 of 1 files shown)


calumbell (approved):
Hi @JeremyMees, thanks for your fix. Issue appears to be solved on my local development server and the test deploy. All tests passing. Nice work!

PR #767, opened 2025-09-22
[BUGFIX] Updated Conditions single-view page to work with updated API reponse

calumbell:
## Build Preview
https://deploy-preview-767--open5e-preview.netlify.app/

## Description

This PR fixes a bug introduced by upcoming breaking changes to the `/v2/conditions` endpoint in the API repository (https://github.com/open5e/open5e-api/pull/794).

The `/conditions/[id]` page now correctly parses the new condition descriptions format.

Tests needed to be updated to account for the changes in the shape of the API response

### DIFF
--- a/app/pages/conditions/[id].vue
+++ b/app/pages/conditions/[id].vue
@@ -11,7 +11,30 @@
         :title="condition.document.name"
       />
     </h1>
-    <md-viewer :text="condition.desc" />
+
+    <!-- Display condition descriptions as a list if API rtns more than one -->
+    <div v-if="condition.descriptions.length > 1">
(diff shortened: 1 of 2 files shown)


eepMoody:
One thought here (which I will leave up to you to decide whether to tackle or leave as an issue for later!) is that like the source selection, we probably want to preference the official WOTC rules in the display order?

It's really a shame they moved away from the "AD&D" brand and let A5e scoop them 😆

eepMoody (approved):

calumbell:
@eepMoody I don't think that is a bad idea at all! My instinct is that this would be better handled on the API end, i'll have a reflect though

PR #773, opened 2025-10-03
Issue 753 Add level annontations to class features

### LINKED ISSUE #753: Add Level annotations to class features on the `/classes/[id]` route
## Description

When viewing the `classes/[id]` route, it would be helpful to see the level that I class ability right next to the ability's name.

This thought was brought on while looking at the 5.2.1 SRD and thinking about how much easier it made the page to navigate.

Image

## Desire Solution

Add an inline level annotation to class feature titles on the `/classes/[id]` page. There are many possible ways that this might be presented:

- Level X: Title
- TItle | Level X
- **Title** *Level X*

Feel free to try out a few options, see which one is the most stylish. My instinct is to go with the third option and make the title and level annotation visually distinct using different TailwindCSS stylings (ie. font-weight an color)

In cases where a feature is gained at multiple levels, only the earliest level should be used as an annotation. ie. For Ability Score Improvements, the Level annotation should be 4

Herissflamm:
## Description

I add a level annotation to the class feature under the name for more visibility
image

## Related Issue

*Closes #753*

## How was this tested?

I opened the Barbarian class page and checked that the feature levels are correct. I also verified that when a feature is gained at multiple levels, the lowest level is shown.

*[Link to a staging deploy](#) page if your item can only be seen on certain pages.*

### DIFF
--- a/app/pages/classes/[className]/index.vue
+++ b/app/pages/classes/[className]/index.vue
@@ -83,7 +83,8 @@
           :id="titleCaseToKebabCase(feature.name)"
           :key="feature.key"
         >
-          <h3>{{ feature.name }}</h3>
+          <h3 class="text-red-400">{{ feature.name }}</h3>
+          <p class="italic text-red-400">{{ "Level " + findFeatureLowestLevel(feature) }}</p>
           <md-viewer
             :text="feature.desc"
             :header-level="3"


eepMoody (approved):
@calumbell this appears to conflict with some of the work you did in yours, want to use your judgement to decide how best to merge?

calumbell:
@eepMoody we can merge both, i'll resolve the conflicts. Thanks for your contribution @Herissflamm ! I am planning to do a pass of the front-end this week and will likely have a heap of new issues that need looking into if you were keen?

PR #785, opened 2025-10-28
[BUGFIX] 2nd Monster in Encounter Builder can now be decremented

### LINKED ISSUE #784: [BUG] Encounter Builder: can only remove/reduce quantity of first monster
Thanks to the user who flagged this issue via the *Report Issue* menu.

## Describe the bug
From the original issue report:
> Encounter Builder. When adding a second or more monsters I can only remove/reduce the first monster in the list.

## Reproduction
From the original issue report:
> Add a monster and a second monsters. You can use `+` to increase any monster but `-` only works on the first monster listed. Once the first monster in the list is gone then the new first monster in the list is the only one that can be reduced.

calumbell:
## Description

This PR fixes an issue flagged by a user via the Report Issue menu where the button for decrementing/removing the second (and only the second) monster on the Encounter Builder didn't do anything when pressed.

Taking a peak at the `useEncounter` composable, there was a weird conditional check that returned if `index === 1`. This doesn't appear to be doing anything apart from breaking this specific button, removing it seemed to have no adverse effect on the component.

## Related Issue

Closes #784

## Test Deploy

https://deploy-preview-785--open5e-preview.netlify.app/monsters

## How was this tested?

- Spot checking on local Nuxt development server (pulling data from a Django development server running the `staging` branch of the Open5e API)
- `npm run test` (all passing)


eepMoody:
I think that one was my bad, a debugging line I threw in while I was stubbing it out and never removed 🤦 Nice catch!

eepMoody (approved):
Ooops, yeah I think that was a stub left over from when i was prototyping things out. Good catch!

calumbell:
Fixed some merge conflicts. Closing and reopening PR to rerun validation

Issue #675, opened 2025-04-02
Copy to Markdown button


calumbell:
## Issue Type

Feature request

## Description

One very common use-case for Open5e is to utilise it in tandem with a word-processor / note-taking app (Obsidian is especially popular), many of which are driven by markdown. With this in mind, being able to quickly take a page and convert it to MD would be a helpful feature.

The idea here is to add a "Copy to Markdown" button to the Open5e website, that when clicked will copy a markdown representation of a page's contents to the user's clipboard. The button itself and all its attendent logic will ideally be contained within a new component, which can then be inserted into `layouts/default.vue` to include it on every page.

This feature would also present a solution to #395, as Homebrewery can also ingest Markdown.

PR #787, opened 2025-10-31
[REFACTOR] Updated `monster/[id]` to preserve white-space when copy-pasting

### LINKED ISSUE #783: [BUG] Copy pasting from `/monsters/[id]` doesn't preserve all white-space / symbol seperators
(Reported via *Report Issue* form. Thanks to the submitter for flagging this!)

## Description

The bug description submitted via the form runs as follows:

> When copying monster stat blocks from a web browser and pasting it into an application that accepts plain text but not HTML, there is missing whitespace which results in the stat block basically being impossible to parse with tools designed to do so, such as the 5e-statblock-importer for FoundryVTT.

My instinct tells me that the issue might be related to using the `tailwindcss` `before:` and `after:` pseudo-selectors to format API data fields to appear in-line: adding space and symbol seperators between fields, that kind of thing.

## To Reproduce

1. Open a new file in your favorite plain text editor (Mac TextEdit in plain text mode or even vi in a command window).
2. Open https://open5e.com/monsters/abbanith-giant in a web browser (I used Firefox).
3. Copy all the text from the heading ""Abbanith Giant"" down to the last separator before the source information.
4. Switch to your editor and paste the clipboard contents.
5. Observe missing whitespace between, for example, ability names and values, or between the word ""Speed"" and ""burrow""."

## Screenshots
Following the reproduction steps here is what my TextEdit window looked like:

Image

calumbell:
## Description

This PR fixes an issue where white-space and symbol seperators were not preserved when copy and pasting text from the `monsters/[id]` page.

The issue was largely caused by the use of Tailwind classes to handle inserting spaces and seperators between listed features (via the `before/after:content-['']` pseudo-selector) which do not interact with copy and pasting. These have been replaced with a little bit of inline Javascript string manipulation, which I think makes much of the markup easier to read.

That the original issue was flagged via the Report Issue form likely indicates that there is a desire for a resolution to #675.

## Related Issue

Closes #783 

## How was this tested?

- Spot checked on local Nuxt development server (`npm run dev`) pulling data from the `staging` branch of the Open5e API, running on a local Django server


eepMoody (approved):

calumbell:
Closing and re-run opening PR to re-run validation

PR #845, opened 2026-03-13
[BUGFIX] `useFindOne` no longer redirects to `/search` when it fails.

### LINKED ISSUE #840: [BUG/UX] Remove redirect to `/search` when `useFindOne` fails
## Description

Currently, when the `useFindOne` composable fails to return any results from the API (ie. the lookup-key doesn't exist in the Open5e dataset), it redirects the user to the `/search`, with the erroneous key as the search query param.

I was the person who implemented this, I thought it would be helpful, and it really isn't. It is actually quite a poor bit of UX for several reasons:
- It doesn't make it clear to the user that the PK they are looking up doesn't exist
- There is no easy way to turn off the redirect on a page-by-page basis, being buried deeply in the composable

## Proposed Solution

Update `composables/api.ts` so that redirects to the `/search` page do not occur when the `useFindOne` composable fails to return any data.

Optionally, add a link on the error page to the `/search` page with the PK used for the failed fetch prefilled as a `search` query parameter. This data should be passed to the error page by throwing it.

calumbell:
## Description

This PR fixes a weird UX interaction on the `useFindOne` composable. On `staging`, when this composable fails to fetch resources (ie. if provided with an invalid key), instead of alerting the user that there is an error it silently redirects them to the `/search` page with the PK of the item that was searched for pre-filled as a query parameter.

The redirect code was located in the `get` functions from `composables/api.ts`. However, refactoring this surfaced an issue with our Vue Query configuration. This was due to the `@hebilicious/vue-query-nuxt` package instantiating the Vue Query plugin once, and then our own custom Vue Query config (`app/plugins/vue-query.ts`) doing so a second time, creating some weird buggy behaviours. These were fixed by uninstalling `@hebilicious/vue-query-nuxt`, so we are only using our custom Vue Query config. I believe this will benefit us long term because, if i am reading Dependabot warnings correctly, `@hebilicious/vue-query-nuxt`'s dependancies were stopping us from being able to patch our Nuxt version.

## Related Issue

Closes #840 

## How was this tested?
- Spot checked on Netlify deploy
- `npm run build` (build process completes without error on local)
- `npm run test` (all tests are passing on local)
- Spot checked on local (pulling data from `staging` branch of API running on a local DJango dev server)

### DIFF
--- a/app/composables/api.ts
+++ b/app/composables/api.ts
@@ -1,9 +1,8 @@
 import { useQuery } from '@tanstack/vue-query';
 import axios from 'axios';
-import { navigateTo, useRoute, useRuntimeConfig } from 'nuxt/app';
+import { useRoute, useRuntimeConfig, type NuxtError } from 'nuxt/app';
 import { computed, unref } from 'vue';
-import { useSourcesList } from './useSourcesList';
-import type { components } from '~/types/open5e-api';
+import type { components } from '@/types/open5e-api';
 
 // Extract schemas from OpenAPI components
 type Schemas = components['schemas'];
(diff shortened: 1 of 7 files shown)


eepMoody (approved):

calumbell:
Merge conflicts fixed. Closing and re-opening to re-run PR validation

Issue #863, opened 2026-03-31
[UX ENHANCEMENT] Add a redirect to `/search` to the Error page


calumbell:
## Description

Currently, on the [beta release](https://beta.open5e.com) of the Open5e website, entering an incorrect key will yield a 404 response from the server, which will trigger a redirect to the `error.vue` page:

<img width="1204" height="776" alt="Image" src="https://github.com/user-attachments/assets/bce73346-de7f-4771-b504-a15deec8d5bb" />

This is intended behaviour, but the UX is a bit awkward. Users are advised to use the search bar to search for what they are looking for instead but there is no direct link to `/search` provided.

A meaningful UX enhancement would be to add a way to link to the search page directly from the error page, Ideally there should be the option of clicking a link to search with the query parameters pre-populated with the key/UID the user had just requested.

For example, if the user was redirected to the error page after attempting to open `/monsters/goblin`, then the URL of the link should be of the form `/search/?text=goblin`.

(Side note: `text` is not a great name for this query parameter. `query` is more idiomatic. But this is outside the scope of this issue)

dan1994:
@calumbell I can take this. A couple of clarifications though:
1. Do you mean the browser address bar? Why is it advised instead of the search bars inside the site?
2. Do we want to have the 404 link to search or redirect to it directly? Assuming the former, would you want something like a "Search for <QUERY> instead" text above the "Return to Homepage" text?

calumbell:
Hi @dan1994 i'll get this assigned to you. Good luck!

w.r.t your points:

1. Not sure I fully understand the question, so apologies if my answer isn't helpful.  Not sure exactly where using the browser bar is implied, but where the search bar this is what is being referred to:

<img width="538" height="331" alt="Image" src="https://github.com/user-attachments/assets/d45fd4d8-58ed-45ea-9d10-06efe3d122e7" />

2. The first one. We want a link and not an immediate redirect. Something like "Search for XYZ" with a link to `/search/?text=XYZ` is what I had in mind.

dan1994:
I'm a bit confused. I tried reproducing this issue, but I'm never redirected to a 404 page when using the search bar. I tried the following queries:
* monsters/goblin
* goblin
* /monsters/goblin
* /monsters/srd_goblin

All but `goblin` provide an empty results search page like this:

<img width="1280" height="1392" alt="Image" src="https://github.com/user-attachments/assets/9adc9a81-4853-4282-a07f-9875e088770d" />

What am I missing?

[...]

PR #867, opened 2026-04-02
fixed bug where useFindPaginated fetched non-existant pages

### LINKED ISSUE #862: [BUG] 404 error when opening empty top-level pages
**Describe the bug**
I observed this bug first on the [Conditions top-level page on the beta deployment](https://beta.open5e.com/conditions). With the default sources selected (`srd-2014` and `srd-2024`) visiting this page will throw a 404 error. 

Image

Investigating the network logs, you can see that there are two requests being made to the `/v2/conditions` endpoint: one for page 1 of the paginated results and the other for page 2. This prefetching is performed to improve the user experience.

It looks like the issue is being caused by the page requesting a second page of results when a the first page is empty.

Inspecting the URLs that are being fetched from, and reproducing their query parameters exactly, can see that:

**Page One** returns a 200 response, but the `results` array is empty

https://api-beta.open5e.com/v2/conditions/?limit=50&page=1&document__key__in=srd-2014,srd-2024&ordering=name&fields=name,key,document&document__fields=name,key&name__contains=

```
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "count": 0,
    "next": null,
    "previous": null,
    "results": []
}
```

**Page two** returns a 404 response.

https://api-beta.open5e.com/v2/conditions/?limit=50&page=2&document__key__in=srd-2014,srd-2024&ordering=name&fields=name,key,document&document__fields=name,key&name__contains=

```
HTTP 404 Not Found
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": "Invalid page."
}
```

I was able to recreate the bug on other top-level pages by only selecting document sources in such a way that I knew that a top-level page would be empty. For example, selecting only the Level Up Advanced 5e Adventurer's Guide (`a5e-ag`) and no other sources, I knew that the API would not return any monster data. I was then able to reproduce the error on the `/monsters` page.

**To Reproduce**
1. Visit the Open5e beta site: https://beta.open5e.com/
2. Ensure that only the default document sources are selected (`srd-2014` and `srd-2024`)
3. Open the network tab in the developers console
4. Click on the Conditions link of the navigation bar. This will take you to the `/conditions` route. You should see a 404 error page after a short time
5. Check the network log and confirm that it was the request for the second page which returned a 404 error code.
...

**Where to begin**
The `useFindPaginated` composable is likely a good place to start looking into to stop this issue. The code that handles prefetching the next page (lines 70-78) are included below:

```
  // Prefetch next page when data changes
  watch(data, () => {
    if (!data.value || pageNo.value === lastPageNo.value ) return;
    const nextPageNo = pageNo.value + 1;
    queryClient.prefetchQuery({
      queryKey: [...queryKey, nextPageNo],
      queryFn: () => findPaginated(createQueryParams(unref(nextPageNo))),
    });
});

```

**Additional Context**

It might be worth considering an additional backend fix so that indexing into pages that do not exist fails more gracefully (a 200 response with an empty list of results, like we got in the above example when we requested page 1)

calumbell:
## Description

This PR fixes an issue with the `useFindPaginated` composable where it would cause a 404 error when attempting to prefetch a page of data where the next page didn't exist.

This was achieved by adding an additional check to the early return of the `watch` function where prefetching is handled:

```
    if (!data.value) return;
    // Early return when there is no next page. This avoids 404 errors
    if (data.value.count === 0 || pageNo.value === lastPageNo.value) return;
```

If there is no `data`, then no prefetching occurs. This appears to fix the problem in all the cases I tested.

## Related Issue

Closes #862 

## How was this tested?
- Spot checked on local Nuxt development server

### DIFF
--- a/app/composables/useFindPaginated.ts
+++ b/app/composables/useFindPaginated.ts
@@ -69,7 +69,9 @@ export function useFindPaginated<T extends keyof EndpointToPaginatedTypeMap>(opt
 
   // Prefetch next page when data changes
   watch(data, () => {
-    if (!data.value || pageNo.value === lastPageNo.value ) return;
+    if (!data.value) return;
+    // Early return when there is no next page. This avoids 404 errors
+    if (data.value.count === 0 || pageNo.value === lastPageNo.value) return;
     const nextPageNo = pageNo.value + 1;
     queryClient.prefetchQuery({
       queryKey: [...queryKey, nextPageNo],


dan1994:
Hi @calumbell, I just opened a [PR](https://github.com/open5e/open5e/pull/868) of my own and saw this. Not sure if there was a miscommunication about me looking into this issue.

calumbell:
Hey @dan1994, this one is on me. I assumed you'd be looking into #863  without checking with you first, apologies. We are due to cut to production on Saturday and wanted to try and get all the truly site-breaking bugs sorted.

I'll go ahead and close this PR and get yours reviewed imminently. It feels like bad form to over-shadow somebody's first contribution to a project!

PR #868, opened 2026-04-02
Fix prefetching of pages when there are no results (#862)

### LINKED ISSUE #862: [BUG] 404 error when opening empty top-level pages

dan1994:
## Description

When opening pages with 0 results, `lastPageNo` was set to 0. This made the condition `pageNo.value === lastPageNo.value` false, thus asking to fetch page 2 which resulted in a 404.
This fix ensures `lastPageNo` is always set to 1 or more, as there is never a page 0.
We also remove the redundant check for `!data.value` which might have originally been there to catch 0 results but should have been `!data.value?.count`.

## Related Issue

Closes #862

## How was this tested?

Manually for now. I wanted to discuss adding tests.

### DIFF
--- a/app/composables/useFindPaginated.ts
+++ b/app/composables/useFindPaginated.ts
@@ -64,18 +64,18 @@ export function useFindPaginated<T extends keyof EndpointToPaginatedTypeMap>(opt
   });
 
   const lastPageNo = computed(() => {
-    return data.value ? Math.ceil(data.value.count / unref(itemsPerPage)) : 1;
+    return data.value?.count ? Math.ceil(data.value.count / unref(itemsPerPage)) : 1;
   });
 
   // Prefetch next page when data changes
   watch(data, () => {
-    if (!data.value || pageNo.value === lastPageNo.value ) return;
+    if (!data.value || pageNo.value === lastPageNo.value) return;
     const nextPageNo = pageNo.value + 1;
     queryClient.prefetchQuery({
       queryKey: [...queryKey, nextPageNo],
       queryFn: () => findPaginated(createQueryParams(unref(nextPageNo))),
     });
-});
+  });
 
   // pagination controls
   const nextPage = () => {


calumbell (changes requested):
Hi @dan1994, thanks for getting this fix made. Works great on the Netlify test deploy and on local Nuxt dev server.

I have two very small changes to request:

1. Could we add the `!data.value` conditional check back? It serves as an unambiguous null check to make sure that `data` exists before attempting to prefetch.

```
    if (!data.value || pageNo.value === lastPageNo.value) return;
```

2. Indentation on line 78 is a bit off

dan1994:
I restored the check, but the formatting change on line 78 is automatic and while I can forcefully ignore it, it seems to be correct as it matches the indentation of the `watch` on line 71.

calumbell (approved):
You are absolutely right about that `});`, completely forgot we were inside a composable. My apologies, thanks for correcting me.

All looks good, thanks for your contribution : )

calumbell:
(Closing and re-opening PR to re-run validation)

PR #920, opened 2026-05-30
Add legacy route redirects and disambiguation

eepMoody:
## Description

Old v1 flat-slug URLs need to reach v2 source-prefixed pages. This could be done with a static mapping table of monsters server-side, but I suspect this approach will be simpler.

A search-based resolution sends unambiguous matches directly to the matching option via 301. and shows a disambiguation page when there are multiple versions and the old link does not include a document suffix.

image

## Refactors and related changes

This substantially refactors the search result construction, primarily by extracting logic used on the search page into helpers and components that can be used elsewhere.

## Testing

Tested locally against production environment. Example links that should work well for testing:

Handle ambiguous link: https://deploy-preview-920--open5e-preview.netlify.app/monsters/zombie
Handle prefixed link: https://deploy-preview-920--open5e-preview.netlify.app/monsters/zombie-bf


calumbell (changes requested):
Thanks for getting this submitted! I think this is really good solution, it plays very well with the existing UX / breadcrumbing!

I have a few, slightly nit-picky changes to requests and we are good to go!:

1. The inner workers of the new components, composables and helpers wasn't immediately obvious to me without a close reading of the source (and I have the context of a PR and plenty of experience in the code-base!). Some comments explaining the purpose and intended use will go a long way.

2.  The call to `useQueryParameter` composable in the `LegacySlugDisambiguation` component feels weird to me. We already extract the query parameters at the page level, so passing the `id` as a prop to the component feels like a much more robust bit of component design.

(I pushed a couple of accessibility and formatting fixes hope you don't mind)

eepMoody:
Actually, now you point it out the whole chain is a bit convoluted. Was pulling some patterns from work that I'm used to, but are probably overkill here. I'm going to take a pass and try and simplify the structure a bit.

eepMoody:
Yeah. I was kinda twisting myself in circles trying to do everything on the page where they landed. I should probably just redirect to a single disambiguation page. Gonna take a fresh pass with that approach instead, it should cut out a lot of the complexity.

calumbell (approved):
Nice, this approach  is a lot cleaner!

Doesn't play quite as nicely with the Breadcrumb links (ie. `Home / Monsters / Zombie / Choice`) but this is a peripheral UX nice-to-have that would be better served in its own issue/PR.

eepMoody:
@calumbell maybe I should just call it "Disambiguation" and it would feel better? 

`monsters/zombie/disambiguation` seems like a pretty normal path for breadcrumb purposes?

PR #923, opened 2026-05-31
update sourcetag color math to produce more variety

eepMoody:
## Description
Source tags for 2014 and 2024 are nearly indistinguishable due to some lazy hashing I set up when I made this originally (the names were all more unique back then)

image

Switched to a better hashing method that produces greater color variety, so the tags are more useful to distinguish the two SRD sources.

image


calumbell (approved):
Definitely an improvement, looks great!

